PR #7, opened 2026-02-11
Add encrypted telemetry support

flip-dots:
The latest firmware on Anker power stations no longer publishes the telemetry unencrypted. This PR fixes that by implementing Ankers bizzare encryption protocol which allows this library to work with power stations running the latest firmware.

There are still a few things that I should probably fix before merging this, as there is something slightly off about the way data is decrypted which prevents some values from being read. More investigation needed.


flip-dots:
I have fixed the way data is decrypted, turns out that the part of the shared secret that is not used as the AES key is used as the IV. It took me half a day to figure that one out, but all telemetry data is now recoverable. There are probably some other changes I will end up making later (e.g add more sensors) but I am happy to merge this for now since the decryption function is done.

PR #8, opened 2026-02-18
Add C300X DC

terratec:
Tested with Firmware v0.1.2.8 but the `UUID_IDENTIFIER` is not advertised (only readable from GATT tree).
Firmware v0.1.3.7 is not working and seems to be paired/bonded without public notify data.

C300X-DC


thomluther:
> Thank you, I will wait.
> Nice find with the databases!

Hi @terratec 
I tried to merge your field findings from [this comment](https://github.com/flip-dots/SolixBLE/pull/8#issue-3959626435) into the MQTT message database in the Api library.
Can you please also install the [Api library](https://github.com/thomluther/anker-solix-api) and use the [mqtt_monitor tool](https://github.com/thomluther/anker-solix-api#mqtt_monitorpy) to verify the merged field descriptions?

There is still some ambiguity about display and light status / mode.
For MQTT messages, I generally use 'display' for values regarding the device display (if existing), and 'light' for values regarding any LED light that devices may optionally provide.
Since BT and MQTT message fields are pretty similar (except the msg header maybe), those findings should be shared across the repos.
The general [message decoding guidelines](https://github.com/thomluther/anker-solix-api/discussions/222) are described there as well.

SolixBLE/devices/c300dc.py
```diff
@@ -246,37 +267,97 @@ def usb_port_a2(self) -> PortStatus:
...
-            bool(self._parse_int("bf", begin=1))
-            if self._data is not None
-            else DEFAULT_METADATA_BOOL
-        )
+        return PortStatus(self._parse_int("bf", begin=1))
 
     @property
-    def dc_output_timeout(self) -> int:
-        """DC output timeout in seconds.
+    def device_overload(self) -> int:
```
flip-dots: Do you know how these overload values work? I would prefer this to be an enum if possible.

flip-dots:
> Should the dependency "Crypto.Cipher" be changed to "Cryptodome.Cipher" ?

If it works sure.

SolixBLE/devices/c300dc.py
flip-dots: Somewhat safe. You will get an exception if you try to read the value and its invalid/unknown but it won't prevent any other values from being read. The advantage of this approach is that it highlights unknown values so when they are encountered there is a decent chance of whoever gets them raising a PR with the correct value, or at least that's the idea anyway.

PR #18, opened 2026-03-15
Improved payload parsing

flip-dots:
I created this from the anker prime branch but this has an additional commit which refactors the payload parsing to provide more useful errors and to handle an edge case discovered [here](https://github.com/flip-dots/HaSolixBLE/issues/3#issuecomment-4063297185) but I am not really happy with my solution yet so I am parking it here.


flip-dots:
This was merged under #16

PR #20, opened 2026-03-20
Codex/c200 minimal

elcritch:
I used Codex/GPT5.3 to do some debugging and setup basic c200 telemetry support. The encryption is different than other models, but adding a unencrypted flag let me read some data.

### DIFF
--- a/SolixBLE/__init__.py
+++ b/SolixBLE/__init__.py
@@ -7,6 +7,7 @@
 from .device import SolixBLEDevice
 from .devices import (
     C300,
+    C200DC,
     C300DC,
     C800,
     C1000,
@@ -31,6 +32,7 @@ from .utilities import discover_devices
 __all__ = [
     "SolixBLEDevice",
     "C300",
+    "C200DC",
     "C300DC",
     "C800",
     "C1000",
--- a/SolixBLE/device.py
+++ b/SolixBLE/device.py
@@ -144,10 +144,6 @@ class SolixBLEDevice:
         _LOGGER.debug(
             f"Established initial connection to '{self.name}' on attempt {self._connection_attempts}!"
         )
-
-        # macOS can report service discovery as incomplete if we try to write
-        # immediately after connect. Ensure services are available first.
-        await self._ensure_services_discovered()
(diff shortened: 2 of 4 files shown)


flip-dots:
@elcritch The changes under #16 have now been merged. It should make it possible to override things like the negotiation and processing methods to bypass decryption. Look at [anker_prime.py](https://github.com/flip-dots/SolixBLE/blob/main/SolixBLE/prime_device.py) to see how thats done and if you need any help let me know :)

flip-dots:
@elcritch Interestingly your device is not the only one which does not do encryption, I have observed that sometimes Anker prime devices (like the 20k power bank) do not use encryption for the telemetry and I merged changes in #33 which allow for parsing unencrypted telemetry packets. In #38 the _TELEMETRY_COMMANDS value was added which allows each device to specify what the telemetry commands for an individual device are, there is an example of it being used for the C1000G2 [here](https://github.com/flip-dots/SolixBLE/blob/af954930155f4ab5366f3e55e4891f558231ee14/SolixBLE/devices/c1000g2.py#L40).

I am happy to modify this code to make it work with the new structure myself if you don't have the time but to do that I could really use some test data so I can be sure my changes will actually work. The data I need would be a log of the packets send/received by the device and the state of the device (e.g USB C1 Power = 10w) and that can be gathered by running examples/monitor.py.

elcritch:
Nice! Yeah unfortunately I won't have time to work on the code. The battery integration was part of a work prototype so I don't have a unit anymore either. We shipped it off somewhere and I didn't even get to keep it. :/ ;)

flip-dots:
> Nice! Yeah unfortunately I won't have time to work on the code. The battery integration was part of a work prototype so I don't have a unit anymore either. We shipped it off somewhere and I didn't even get to keep it. :/ ;)

Fair enough, I think I will close this PR for now but I might later use some of the mappings from the code to add experimental support for the C200 at a later date and then wait and see if anyone raises any issues. Regardless thanks for your help, I appreciate all the effort that others have put in to help make this library work better.

Issue #28, opened 2026-03-30
Testing on Solarbank 2 E1600 Pro


jul1an-s:
Hi,

first off: Amazing project, thank you so much for your work! 🎉

I'm trying to get the Solarbank 2 E1600 Pro working with this project. Ultimate goal would be to control the output power through the schedule like @smariacher did with the Solarbank 1 https://github.com/flip-dots/SolixBLE/issues/21

I got some first telemetry dumps, but apparently it can't decrypt the telemetry data properly. 

```

DEBUG:bleak_retry_connector:F4:9D:8A:13:CE:82 - F4:9D:8A:13:CE:82: Connected after 1 attempts
DEBUG:SolixBLE.device:Established initial connection to 'Solarbank 2 E1600 Pro' on attempt 1!
DEBUG:SolixBLE.device:Subscribing to notifications from device 'Solarbank 2 E1600 Pro'!
DEBUG:bleak.backends.bluezdbus.client:using "AcquireNotify" for notifications on characteristic 17
DEBUG:bleak.backends.bluezdbus.manager:received D-Bus signal: org.freedesktop.DBus.Properties.PropertiesChanged (/org/bluez/hci0/dev_F4_9D_8A_13_CE_82/service000e/char0011): ['org.bluez.GattCharacteristic1', {'NotifyAcquired': <dbus_fast.signature.Variant ('b', True)>}, []]
DEBUG:SolixBLE.device:Sending negotiation initiation request to 'Solarbank 2 E1600 Pro'...
DEBUG:SolixBLE.device:The client the notification is from is: BleakClient, F4:9D:8A:13:CE:82
DEBUG:SolixBLE.device:Received notification from 'Solarbank 2 E1600 Pro'. length: 14, packet: 'ff090e00030001080100a1010152'
DEBUG:SolixBLE.device:Pattern: 030001
DEBUG:SolixBLE.device:CMD: 0801
DEBUG:SolixBLE.device:Payload: 00a10101
DEBUG:SolixBLE.device:Payload length: 4
DEBUG:SolixBLE.device:Received encryption negotiation message!
DEBUG:SolixBLE.device:Entered negotiation stage 1 due to response from device!
DEBUG:SolixBLE.device:Parameters: {'a1': '01'}
DEBUG:SolixBLE.device:Sending stage 1 response message...
[...]

PR #32, opened 2026-04-12
Improve frida script logging

flip-dots:
This PR contains changes to the Frida script to attempt to get it to log encryption data that seems to be missing in the logs captured by @jul1an-s under #28.


flip-dots:
This did not end up fixing the original issue but it is still probably useful so I think it makes sense to merge it.

PR #33, opened 2026-04-22
Fix intermittent protocol issues on Anker Prime 160w charger.

flip-dots:
An [issue](https://github.com/flip-dots/HaSolixBLE/issues/24) was raised on the Home Assistant integration page for issues with the Anker Prime 250w charger but issues with the 160w charger were also reported there and the logs from it are quite strange. The logs show that the charger is sending unencrypted telemetry data for some reason. What is especially weird is that my unit does not do this despite running the same firmware version (1.6.0.6). 

At the moment this PR adds support for reading unencrypted telemetry data but more may be added to deal with other strange things from the logs.

### DIFF
--- a/SolixBLE/device.py
+++ b/SolixBLE/device.py
@@ -518,7 +518,9 @@ class SolixBLEDevice:
         )
         return cipher.encrypt(padded_data)
 
-    async def _process_telemetry_packet(self, payload: bytes, cmd: bytes = None) -> None:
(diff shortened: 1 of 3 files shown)


flip-dots:
I have recently obtained an Anker prime 220w 20k power bank and it also exhibits the same thing where some telemetry data is sent unencrypted, so this change also allows for it to work. Adding actual support for it will be done under a different PR.

Issue #34, opened 2026-04-23
F2600


unex:
Hello, I have an F2600 and have been working on implementing it for use in HaSolixBLE.  I have found that it is essentially a copy of the existing F2000 implementation, even having the same output ports, so I'm not sure if it makes sense to have a complete duplicate or if there should be a "base" model that the F200 and F2600 use.  If I add it in HaSolixBLE as a F2000 it appears to work perfectly minus having to restart HA to get it to reconnect but that is probably unrelated.

I also got a modified version of the "patch" script working and verified that the commands are the same as on the C1000,  plus mapped out commands for the other features I could find like charging power and power saving mode.

flip-dots:
@unex nice, do you plan on submitting a PR or would you rather I add a dedicated class for it using the info you have provided?

If you want to implement it yourself I am thinking we make a new F2600 class which extends the F2000 class but we put the control functions in the F2600 class rather than the F2000 class out of an abundance of caution, since we do not know that they are the same (even though they probably are) and sending commands to the device that we do not know for certain are safe probably isn't a great idea, though I don't mind doing that for parsing the telemetry values since the worse that can happen is some wrong values. 

In theory you should not have to restart HA to get it to reconnect and with the devices I have available to test (C300, C1000, Prime 160w charger) that is not necessary, but these Anker devices tend to behave strangely, so I am not all that surprised.

unex:
@flip-dots I can make a PR, just wanted to double check how you wanted it integrated, I will make it extend F2000.

PR #35, opened 2026-04-29
Add F2600 support

unex:
Add F2600 support as discussed in #34 

Tested on my own F2600 to the best of my ability.

It seems that some parameters like AC charge power and display timeout seconds are only returned during a full "get_status_update" response.  Currently the `_process_telemetry` function will overwrite `_data` on every call, so if there is a response from the battery that does not include everything, all other parameters will be removed.  While I was doing my testing I made a change to update parameters instead of a full overwrite, if you think that is useful it can be found here: https://github.com/unex/SolixBLE/tree/update-parameters


SolixBLE/devices/f2600.py
```diff
@@ -0,0 +1,321 @@
...
+PAYLOAD_AC_CHARGING_POWER = "a10121a20302"
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class F2600(F2000):
+    """
+    F2600 Power Station.
+
+    Use this class to connect and monitor a F2600 power station.
```
flip-dots, line 44: This could also use the model number like in the [other](https://github.com/flip-dots/SolixBLE/blob/18849c92ff10125fc9291379603ad4b67eeab788/SolixBLE/devices/c1000.py#L34) device classes. This [manual](https://www.manualslib.com/manual/3246830/Anker-Solix-F2600.html) seems to suggest its A1781.

SolixBLE/devices/f2600.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        return PortStatus(self._parse_int("c6", begin=1))
+
+    @property
+    def usb_c2(self) -> PortStatus:
+        """USB C2 Port Status.
+
+        :returns: Status of the USB C2 port.
+        """
+        return PortStatus(self._parse_int("c7", begin=1))
```
flip-dots, line 157: Needs to use the same naming scheme as other devices (e.g [c300](https://github.com/flip-dots/SolixBLE/blob/18849c92ff10125fc9291379603ad4b67eeab788/SolixBLE/devices/c300.py#L264-L270))

SolixBLE/devices/f2600.py
```diff
@@ -0,0 +1,321 @@
...
+
+    async def set_ac_charging_power(self, watts: int) -> None:
+        """Set the AC charging power limit in watts.
+
+        :param watts: AC charging power limit in watts.
+        :raises ValueError: If power value is out of valid range.
+        :raises ConnectionError: If not connected to device.
+        :raises BleakError: If command transmission fails.
+        """
+        if watts < 100 or watts > 1440: # below 100 causes max charge, 1440 is max in app.
```
flip-dots: Does it let you set arbitrary values?, in my experience these things are limited to specific values like 100, 200, 300, etc.

SolixBLE/devices/f2600.py
```diff
@@ -0,0 +1,321 @@
...
+
+import logging
+
+from ..const import (
+    DEFAULT_METADATA_BOOL,
+    DEFAULT_METADATA_INT,
+)
+from ..states import DisplayTimeout, LightStatus, PortStatus
+
+from . import F2000
```
flip-dots, line 17: Since this is a new device a page needs to be added to the docs for it as well as updating the index page with the supported capabilities, #14 should be a good example of how to do it.

flip-dots (changes requested):
So this is pretty good, just a few minor things. 

The whole get_status_update thing is not very well thought out and that is my fault, I added it as a "this is kind of neat but I don't have a use for it" type of thing and did not look into it any further and left it kind of half arsed. It's something I will probably address eventually but these kinds of architectural changes will need to wait until I have more time when my dissertation and exams are over which will be in a couple of months. 

Since this is a fairly simple PR which does not require much time on my part if you just fix the mentioned issues from the comments I am happy to merge it and deal with the bigger architectural changes later, though I might not have the time to add it to the Home Assistant integration for a while.

SolixBLE/devices/f2600.py
unex: :+1:

SolixBLE/devices/f2600.py
unex: :+1:

SolixBLE/devices/f2600.py
unex: From my testing it accepts anything between 100 and 1440, outside of this range it just get set to 1440

unex:
I've done some more testing and found some discrepancies from the F2000 implementation:

- `af` will show DC input power as well, I believe this is total input power, even though from what I can tell it wont use both solar input and AC input at the same time.
- `b0` is combined output, not just AC
- `a5` appears to be AC input power, I cant find a key that correlates to `ac_to_battery` specifically, and it seems that the device only reports AC input power when the battery is charging.  Once the battery is charged all AC input values go to 0 even though the device is doing passthrough.

I am not sure what the difference is between AC_POWER_OUT and AC_POWER_OUT_SOCKETS is, since these models only have AC sockets.

SolixBLE/devices/f2600.py
unex: :+1:

flip-dots:
Hi,

This looks fine but I don't currently have the time to do a proper review and release, I still have a few weeks of exams left but I should be able to sort this out after that, sorry :(

flip-dots:
So I have finally had the time to do a review and it looks good but it could use some tests. I would like to see some tests for the values like the ones [here](https://github.com/flip-dots/SolixBLE/blob/7c14feb7280a75719bd800f866e36ca1904a9a52/tests/test_devices.py#L822) and maybe some for the commands like the ones [here](https://github.com/flip-dots/SolixBLE/blob/7c14feb7280a75719bd800f866e36ca1904a9a52/tests/test_devices.py#L842-L843) since I plan on doing quite a bit of refactoring soon and would rather not accidentally break this stuff since I don’t have this device to test with. They don't need to be perfect since the existing tests are in need of some major refactoring but it would be good to have something. You can get test coverage using pytest --cov which should help guide you on anything you might have missed.

 I am not going to require you to do it, but it would also be great to have a test added for the [negotiation](https://github.com/flip-dots/SolixBLE/blob/7c14feb7280a75719bd800f866e36ca1904a9a52/tests/test_devices.py#L953) using data captured from your device since these do vary across devices and it would help me in future if I ever get around to properly figuring out how negotiation works. 

Other than all that this looks great, happy to merge once the merge conflict is resolved and some tests are added.

unex:
Sorry forgot about this for a bit, I have added tests and fixed the merge conflict.  

Let me know if you need anything else.

SolixBLE/devices/f2600.py
```diff
@@ -0,0 +1,490 @@
...
+    @property
+    def solar_port(self) -> PortStatus:
+        """Solar/DC input port status.
+
+        Note: remains INPUT after the Anderson connector loses power until
+        AC wall charging takes over, at which point it clears to NOT_CONNECTED.
+
+        :returns: Status of the solar/DC input port.
+        """
+        return (
```
flip-dots, line 210: It should be possible to use [from_input_only](https://github.com/flip-dots/SolixBLE/blob/cb684b333467e7b1582509a67770223af549564a/SolixBLE/states.py#L26) like [here](https://github.com/flip-dots/SolixBLE/blob/cb684b333467e7b1582509a67770223af549564a/SolixBLE/devices/c300dc.py#L388) instead of needing this if statement.

SolixBLE/devices/f2600.py
```diff
@@ -0,0 +1,490 @@
...
+
+        :returns: Total output power (AC + USB + DC) in watts or default int value.
+        """
+        return self._parse_int("b0", begin=1)
+
+    @property
+    def solar_port(self) -> PortStatus:
+        """Solar/DC input port status.
+
+        Note: remains INPUT after the Anderson connector loses power until
```
flip-dots: You need to use [this format](https://github.com/flip-dots/SolixBLE/blob/cb684b333467e7b1582509a67770223af549564a/SolixBLE/devices/f2000.py#L24) for notes to work properly in the documentation.

SolixBLE/devices/f2600.py
```diff
@@ -0,0 +1,490 @@
...
+
+        :returns: Total input power in watts or default int value.
+        """
+        return self._parse_int("af", begin=1)
+
+    @property
+    def ac_power_in(self) -> int:
+        """AC Power In.
+
+        On F2600, key ``a5`` tracks total AC wall input. Key ``af`` tracks combined total of all
```
flip-dots: Use the collapse tag like [here](https://github.com/flip-dots/SolixBLE/blob/cb684b333467e7b1582509a67770223af549564a/SolixBLE/devices/maggo_3in1.py#L27) to hide technical details from someone reading the documentation which are not needed to simply use the library. So things like parameter mappings should be hidden but other things like the note about the solar input should not be hidden.

SolixBLE/devices/f2600.py
```diff
@@ -0,0 +1,490 @@
...
+        return self._parse_int("a5", begin=1)
+
+    @property
+    def ac_charging_power(self) -> int:
+        """Configured AC charging power limit in watts.
+
+        :returns: AC charging power limit or default int value.
+        """
+        if self._data is None or "d1" not in self._data:
+            return DEFAULT_METADATA_INT
```
flip-dots, line 245: Is this check needed?

SolixBLE/devices/f2600.py
```diff
@@ -0,0 +1,490 @@
...
+        return self._parse_int("d1", begin=1)
+
+    @property
+    def display_timeout_seconds(self) -> int:
```
flip-dots: Where possible use existing names for properties, in this case [display_timeout](https://github.com/flip-dots/SolixBLE/blob/cb684b333467e7b1582509a67770223af549564a/SolixBLE/devices/f3800.py#L328) already exists and is appropriate.

SolixBLE/devices/f2600.py
```diff
@@ -0,0 +1,490 @@
...
+        """Configured display timeout in seconds.
+
+        :returns: Display timeout in seconds or default int value.
+        """
+        if self._data is None or "d3" not in self._data:
```
flip-dots, line 254: Is this needed? Same as the other instances where it is used. It should only be needed when you aren't using a helper like _parse_int, _parse_string, etc or when you are changing its type e.g converting an int to a float like [here](https://github.com/flip-dots/SolixBLE/blob/cb684b333467e7b1582509a67770223af549564a/SolixBLE/devices/prime_charger_250w.py#L53-L62).

flip-dots (commented):
Nice work with the tests, there are just a few minor things I noticed when skimming the PR, there is also a merge conflict in the docs because #56 was merged which changed the device support page. 

Should be good to merge after those things are fixed.

SolixBLE/devices/f2600.py
unex: I don't think from_input_only works here, bf isn't a PortStatus on this model. from_input_only only remaps 1 → 2 and passes a raw 2 straight through as INPUT, but the F2600 reports bf = 2 while AC wall charging with nothing connected to the Anderson port. from_input_only would report solar input where there is none.

SolixBLE/devices/f2600.py
unex: If I remember correctly _parse_int only guards self._data is None and then does self._data[key], which raises KeyError when the key is absent. d1, d3, d9, db and de are only present in a full get_status_update response so an unsolicited telemetry packet doesn't carry them since _process_telemetry clears _data

SolixBLE/devices/f2600.py
unex: same as https://github.com/flip-dots/SolixBLE/pull/35#discussion_r3670602676

Issue #22, opened 2026-03-22
Support for Solix C1000 Gen 2?


crowedavid:
i have been following the incredible progress here in adding support for additional models and decided to see if my Solix C1000 Gen 2 would be recognized and how well it would show up - unfortunately not very much (used SolixBLE v3.4.0 along with the example [monitoring](https://solixble.readthedocs.io/en/latest/examples.html#monitoring-example) script).

intent is to use the Home Assistant HaSolixBLE integration but wanted to gather data direct w/o disrupting my HA instance too much. i have had this same C1000G2 setup using the Anker Solix HA integration with data updates working well on wifi. hopefully the data format/fields will come over easily as it appears they did for other devices.

C1000G2 is using the latest firmware for it: v1.1.4.1

here is the plain state output from the monitor.py script: (full debug output attached)

==== REMOTE STATE CHANGE DETECTED ====
C1000G2(
    AC_OUTPUT: PortStatus.UNKNOWN,
    AC_POWER_IN: -1,
    AC_POWER_OUT: -1,
    ADDRESS: 7C:E9:13:4E:C8:62,
    AVAILABLE: False,
    BATTERY_HEALTH: -1,
    BATTERY_PERCENTAGE: -1,
    CONNECTED: True,
    DC_OUTPUT: PortStatus.UNKNOWN,
    DC_POWER_OUT: -1,
    LAST_UPDATE: None,
    MAX_BATTERY_PERCENTAGE: -1,
    MIN_BATTERY_PERCENTAGE: -1,
    NAME: SOLIX C1000 Gen 2,
    NEGOTIATED: True,
    PART_NUMBER: Unknown,
    POWER_OUT: -1,
    SERIAL_NUMBER: Unknown,
    SOLAR_PORT: PortStatus.UNKNOWN,
    SOLAR_POWER_IN: -1,
    TEMPERATURE: -1,
    USB_A1_POWER: -1,
    USB_C1_POWER: -1,
    USB_C2_POWER: -1,
    USB_C3_POWER: -1,
    USB_PORT_A1: PortStatus.UNKNOWN,
    USB_PORT_C1: PortStatus.UNKNOWN,
    USB_PORT_C2: PortStatus.UNKNOWN,
    USB_PORT_C3: PortStatus.UNKNOWN,
)

[solix.txt](https://github.com/user-attachments/files/26169513/solix.txt)

does this help in getting the information needed?  if not, i'm happy to gather more data and try out updates on my device.

[...]

PR #36, opened 2026-05-20
Solarbank 2 Schedule/Power control

jul1an-s:
### Changes:

- Added Prime style handshake to SB2
- Controls:
   - Added the code to send simple schedules to SB2 `405e`. Currently restricted to setting one power output 24/7
   - Added boilerplate for additional Commands `[untested]`:
      - light control
      - Added Reserved Power control
      - Added max output power control
- Will add a frida script for dart lib as well
- Extended gitignore

### Tests & Docu
- Docu updated
- Added a few new tests
- Tests passed

### Why two handshakes?
As I mentioned in the issue #28 I have implemented both handshakes/protocol variants:
##### Prime handshake
I included the Prime style variant because it seems like the "right" way to interact with the device and a lot of work went into it. The legacy handshake might get removed any time, so it's good to have a fallback option.

##### "Legacy" handshake
I included the Legacy variant because right now it is so much simpler for our use case. No env vars, no user specific config.

### DIFF
--- a/SolixBLE/devices/solarbank2.py
+++ b/SolixBLE/devices/solarbank2.py
@@ -235,6 +235,12 @@ class Solarbank2Common(SolixBLEDevice):
             cmd=bytes.fromhex(CMD_SB2_SET_SCHEDULE), payload=payload
         )
 
+    ##############################
+    #          WARNING           #
+    #   Functions below are      #
+    # NOT E2E Tested on Device!! #
+    ##############################
+
     @staticmethod
     def _build_set_light_payload(light_on: bool) -> bytes:
         """Build the plaintext payload for cmd 0x4068 (status light on/off).
@@ -245,11 +251,19 @@ class Solarbank2Common(SolixBLEDevice):
             a1 01 21              command marker
             a2 02 01 00           constant 0x00 flag (purpose unknown)
             a3 02 01 <state>      0 = ON, 1 = OFF (the "light_off_switch" bit)
+
+        .. warning::
+            This function has been added using telemtry captures. It has not 
+            been tested!
         """
         return bytes.fromhex(f"a10121a2020100a30201{0 if light_on else 1:02x}")
 
     async def set_light_switch(self, light_on: bool) -> None:
         """Turn the SB2 status light on or off.
+
+        .. warning::
+            This function has been added using telemtry captures. It has not 
+            been tested!
 
         :param light_on: ``True`` to turn the light on, ``False`` to turn it off.
         :raises ConnectionError: If not connected/negotiated to the device.
@@ -278,6 +292,10 @@ class Solarbank2Common(SolixBLEDevice):
             a2 02 01 <pct>     -> c405 telemetry b4 (discharge cutoff)
             a3 02 01 <b5>      -> c405 telemetry b5 (low-power-input parameter)
             a4 02 01 <pct>     -> c405 telemetry b6 (charge cutoff)
+
+        .. warning::
+            This function has been added using telemtry captures. It has not 
+            been tested!
         """
         if level is SBPowerCutoff.UNKNOWN:
             raise ValueError("SBPowerCutoff.UNKNOWN is not a valid setter input")
@@ -297,6 +315,10 @@ class Solarbank2Common(SolixBLEDevice):
         Mirrors the Anker app's "Reserved power" toggle. Writes three
         telemetry fields (output_cutoff_data, lowpower_input_data,
         input_cutoff_data).
+
+        .. warning::
+            This function has been added using telemtry captures. It has not 
+            been tested!
 
         :param level: Desired reserved-power level
             (:class:`SBPowerCutoff.P5` or :class:`SBPowerCutoff.P10`).
@@ -318,6 +340,10 @@ class Solarbank2Common(SolixBLEDevice):
             a1 01 21
             a2 03 02 <u16 LE watts>     output power limit
             a3 03 02 00 00              constant zero (flags/secondary cap?)
+
+        .. warning::
+            This function has been added using telemtry captures. It has not 
+            been tested!
         """
         if load is MaxLoadSB2.UNKNOWN:
             raise ValueError("MaxLoadSB2.UNKNOWN is not a valid setter input")
@@ -327,6 +353,10 @@ class Solarbank2Common(SolixBLEDevice):
     async def set_max_load(self, load: MaxLoadSB2) -> None:
         """Set the AC output power limit (max load) in watts.
 
+        .. warning::
+            This function has been added using telemtry captures. It has not 
+            been tested!
+
         :param load: One of the discrete limits in :class:`MaxLoadSB2`.
         :raises ConnectionError: If not connected/negotiated to the device.
         :raises ValueError: If ``load`` is ``MaxLoadSB2.UNKNOWN``.
@@ -335,6 +365,11 @@ class Solarbank2Common(SolixBLEDevice):
         await self._send_command(
             cmd=bytes.fromhex(CMD_SB2_SET_MAX_LOAD), payload=payload
         )
+
+    ##############################
+    #        END UNTESTED        #
+    #          FUNCTIONs         #
+    ##############################
 
     # Telemetry property accessors
 
--- a/SolixBLE/devices/solarbank2.py
+++ b/SolixBLE/devices/solarbank2.py
@@ -77,23 +77,6 @@ class MaxLoadSB2(Enum):
     W1000 = 1000
 
 
-class LightSwitchSB2(Enum):
-    """Solarbank 2 status-light switch state (setter input for cmd ``0x4068``).
-
-    Values match the underlying ``light_off_switch`` bit on the wire (which
-    the ``light_on`` telemetry property inverts for a user-friendly bool).
-    """
-
-    #: Unknown state.
-    UNKNOWN = -1
-
-    #: Light is on.
-    ON = 0
-
-    #: Light is off.
-    OFF = 1
-
-
 #: One of the command codes for setting a schedule on an SB2.
 CMD_SB2_SET_SCHEDULE = "405e"
 
@@ -253,7 +236,7 @@ class Solarbank2Common(SolixBLEDevice):
         )
 
     @staticmethod
-    def _build_set_light_payload(state: LightSwitchSB2) -> bytes:
+    def _build_set_light_payload(light_on: bool) -> bytes:
         """Build the plaintext payload for cmd 0x4068 (status light on/off).
 
         Plaintext layout (without the ``fe 05 03 <ts>`` trailer added by
@@ -263,18 +246,15 @@ class Solarbank2Common(SolixBLEDevice):
             a2 02 01 00           constant 0x00 flag (purpose unknown)
             a3 02 01 <state>      0 = ON, 1 = OFF (the "light_off_switch" bit)
         """
-        if state is LightSwitchSB2.UNKNOWN:
-            raise ValueError("LightSwitchSB2.UNKNOWN is not a valid setter input")
-        return bytes.fromhex(f"a10121a2020100a30201{state.value:02x}")
+        return bytes.fromhex(f"a10121a2020100a30201{0 if light_on else 1:02x}")
 
-    async def set_light_switch(self, state: LightSwitchSB2) -> None:
+    async def set_light_switch(self, light_on: bool) -> None:
         """Turn the SB2 status light on or off.
 
-        :param state: Desired light state.
+        :param light_on: ``True`` to turn the light on, ``False`` to turn it off.
         :raises ConnectionError: If not connected/negotiated to the device.
-        :raises ValueError: If ``state`` is ``LightSwitchSB2.UNKNOWN``.
         """
-        payload = self._build_set_light_payload(state)
+        payload = self._build_set_light_payload(light_on)
         await self._send_command(
             cmd=bytes.fromhex(CMD_SB2_SET_LIGHT), payload=payload
         )
--- a/SolixBLE/devices/solarbank2.py
+++ b/SolixBLE/devices/solarbank2.py
@@ -77,8 +77,35 @@ class MaxLoadSB2(Enum):
     W1000 = 1000
 
 
+class LightSwitchSB2(Enum):
+    """Solarbank 2 status-light switch state (setter input for cmd ``0x4068``).
+
+    Values match the underlying ``light_off_switch`` bit on the wire (which
+    the ``light_on`` telemetry property inverts for a user-friendly bool).
+    """
+
+    #: Unknown state.
+    UNKNOWN = -1
+
+    #: Light is on.
+    ON = 0
+
+    #: Light is off.
+    OFF = 1
+
+
 #: One of the command codes for setting a schedule on an SB2.
 CMD_SB2_SET_SCHEDULE = "405e"
+
+#: Command code for setting the AC output power limit (max load).
+CMD_SB2_SET_MAX_LOAD = "4080"
+
+#: Command code for setting the reserved-power composite (discharge cutoff,
+#: low-power-input parameter, and charge cutoff written together).
+CMD_SB2_SET_RESERVED_POWER = "4067"
+
+#: Command code for turning the status light on or off.
+CMD_SB2_SET_LIGHT = "4068"
 
 #: TLV header introducing a 4-byte LE timestamp (tag ``a1``, length ``04``).
 #: Recurs across handshake stage plaintexts and post-handshake payloads.
@@ -223,6 +250,110 @@ class Solarbank2Common(SolixBLEDevice):
         payload = self._build_set_schedule_payload(power_w)
         await self._send_command(
             cmd=bytes.fromhex(CMD_SB2_SET_SCHEDULE), payload=payload
+        )
+
+    @staticmethod
+    def _build_set_light_payload(state: LightSwitchSB2) -> bytes:
+        """Build the plaintext payload for cmd 0x4068 (status light on/off).
+
+        Plaintext layout (without the ``fe 05 03 <ts>`` trailer added by
+        ``_send_command``)::
+
+            a1 01 21              command marker
+            a2 02 01 00           constant 0x00 flag (purpose unknown)
+            a3 02 01 <state>      0 = ON, 1 = OFF (the "light_off_switch" bit)
+        """
+        if state is LightSwitchSB2.UNKNOWN:
+            raise ValueError("LightSwitchSB2.UNKNOWN is not a valid setter input")
+        return bytes.fromhex(f"a10121a2020100a30201{state.value:02x}")
+
+    async def set_light_switch(self, state: LightSwitchSB2) -> None:
+        """Turn the SB2 status light on or off.
+
+        :param state: Desired light state.
+        :raises ConnectionError: If not connected/negotiated to the device.
+        :raises ValueError: If ``state`` is ``LightSwitchSB2.UNKNOWN``.
+        """
+        payload = self._build_set_light_payload(state)
+        await self._send_command(
+            cmd=bytes.fromhex(CMD_SB2_SET_LIGHT), payload=payload
+        )
+
+    # Mapping for 3rd field reserved power command
+    _RESERVED_POWER_A3_MAP: dict[int, int] = {5: 4, 10: 5}
+
+    @staticmethod
+    def _build_set_reserved_power_payload(level: SBPowerCutoff) -> bytes:
+        """Build the plaintext payload for cmd 0x4067 (reserved power).
+
+        The Anker app's "Reserved power" writes 
+        three independent fields. This is reflected in the telemtry:
+        discharge cutoff (b4), low-power-input (b5), and charge cutoff (b6).
+        a2 and a4 both equal the reserved-power %; a3 seems to follow an
+        uncharacterized mapping.
+
+        Plaintext layout (without the ``fe 05 03 <ts>`` trailer)::
+
+            a1 01 21
+            a2 02 01 <pct>     -> c405 telemetry b4 (discharge cutoff)
+            a3 02 01 <b5>      -> c405 telemetry b5 (low-power-input parameter)
+            a4 02 01 <pct>     -> c405 telemetry b6 (charge cutoff)
+        """
+        if level is SBPowerCutoff.UNKNOWN:
+            raise ValueError("SBPowerCutoff.UNKNOWN is not a valid setter input")
+        pct = level.value
+        a3_value = Solarbank2Common._RESERVED_POWER_A3_MAP.get(pct)
+        if a3_value is None:
+            raise ValueError(
+                f"Reserved-power value {pct}% has no captured a3 mapping."
+            )
+        return bytes.fromhex(
+            f"a10121a20201{pct:02x}a30201{a3_value:02x}a40201{pct:02x}"
+        )
+
+    async def set_reserved_power(self, level: SBPowerCutoff) -> None:
+        """Set the battery reserved-power percentage.
+
+        Mirrors the Anker app's "Reserved power" toggle. Writes three
+        telemetry fields (output_cutoff_data, lowpower_input_data,
+        input_cutoff_data).
+
+        :param level: Desired reserved-power level
+            (:class:`SBPowerCutoff.P5` or :class:`SBPowerCutoff.P10`).
+        :raises ConnectionError: If not connected/negotiated to the device.
+        :raises ValueError: If ``level`` is ``UNKNOWN`` or has no captured
+            a3 mapping.
+        """
+        payload = self._build_set_reserved_power_payload(level)
+        await self._send_command(
+            cmd=bytes.fromhex(CMD_SB2_SET_RESERVED_POWER), payload=payload
+        )
+
+    @staticmethod
+    def _build_set_max_load_payload(load: MaxLoadSB2) -> bytes:
+        """Build the plaintext payload for cmd 0x4080 (AC power limit).
+
+        Plaintext layout (without the ``fe 05 03 <ts>`` trailer)::
+
+            a1 01 21
+            a2 03 02 <u16 LE watts>     output power limit
+            a3 03 02 00 00              constant zero (flags/secondary cap?)
+        """
+        if load is MaxLoadSB2.UNKNOWN:
+            raise ValueError("MaxLoadSB2.UNKNOWN is not a valid setter input")
+        watts_le_hex = load.value.to_bytes(2, "little").hex()
+        return bytes.fromhex(f"a10121a20302{watts_le_hex}a303020000")
+
+    async def set_max_load(self, load: MaxLoadSB2) -> None:
+        """Set the AC output power limit (max load) in watts.
+
+        :param load: One of the discrete limits in :class:`MaxLoadSB2`.
+        :raises ConnectionError: If not connected/negotiated to the device.
+        :raises ValueError: If ``load`` is ``MaxLoadSB2.UNKNOWN``.
+        """
+        payload = self._build_set_max_load_payload(load)
+        await self._send_command(
+            cmd=bytes.fromhex(CMD_SB2_SET_MAX_LOAD), payload=payload
         )
 
     # Telemetry property accessors
--- a/SolixBLE/devices/solarbank2.py
+++ b/SolixBLE/devices/solarbank2.py
@@ -241,12 +241,6 @@ class Solarbank2Common(SolixBLEDevice):
             cmd=bytes.fromhex(CMD_SB2_SET_SCHEDULE), payload=payload
         )
 
-    ##############################
-    #          WARNING           #
-    #   Functions below are      #
-    # NOT E2E Tested on Device!! #
-    ##############################
-
     @staticmethod
     def _build_set_light_payload(light_on: bool) -> bytes:
         """Build the plaintext payload for cmd 0x4068 (status light on/off).
@@ -257,19 +251,11 @@ class Solarbank2Common(SolixBLEDevice):
             a1 01 21              command marker
             a2 02 01 00           constant 0x00 flag (purpose unknown)
             a3 02 01 <state>      0 = ON, 1 = OFF (the "light_off_switch" bit)
-
-        .. warning::
-            This function has been added using telemetry captures. It has not 
-            been tested!
         """
         return bytes.fromhex(f"a10121a2020100a30201{0 if light_on else 1:02x}")
 
     async def set_light_switch(self, light_on: bool) -> None:
         """Turn the SB2 status light on or off.
-
-        .. warning::
-            This function has been added using telemetry captures. It has not 
-            been tested!
 
         :param light_on: ``True`` to turn the light on, ``False`` to turn it off.
         :raises ConnectionError: If not connected/negotiated to the device.
@@ -298,10 +284,6 @@ class Solarbank2Common(SolixBLEDevice):
             a2 02 01 <pct>     -> c405 telemetry b4 (discharge cutoff)
             a3 02 01 <b5>      -> c405 telemetry b5 (low-power-input parameter)
             a4 02 01 <pct>     -> c405 telemetry b6 (charge cutoff)
-
-        .. warning::
-            This function has been added using telemetry captures. It has not 
-            been tested!
         """
         if level is SBPowerCutoff.UNKNOWN:
             raise ValueError("SBPowerCutoff.UNKNOWN is not a valid setter input")
@@ -321,10 +303,6 @@ class Solarbank2Common(SolixBLEDevice):
         Mirrors the Anker app's "Reserved power" toggle. Writes three
         telemetry fields (output_cutoff_data, lowpower_input_data,
         input_cutoff_data).
-
-        .. warning::
-            This function has been added using telemetry captures. It has not 
-            been tested!
 
         :param level: Desired reserved-power level
             (:class:`SBPowerCutoff.P5` or :class:`SBPowerCutoff.P10`).
@@ -346,10 +324,6 @@ class Solarbank2Common(SolixBLEDevice):
             a1 01 21
             a2 03 02 <u16 LE watts>     output power limit
             a3 03 02 00 00              constant zero (flags/secondary cap?)
-
-        .. warning::
-            This function has been added using telemetry captures. It has not 
-            been tested!
         """
         if load is MaxLoadSB2.UNKNOWN:
             raise ValueError("MaxLoadSB2.UNKNOWN is not a valid setter input")
@@ -359,10 +333,6 @@ class Solarbank2Common(SolixBLEDevice):
     async def set_max_load(self, load: MaxLoadSB2) -> None:
         """Set the AC output power limit (max load) in watts.
 
-        .. warning::
-            This function has been added using telemetry captures. It has not 
-            been tested!
-
         :param load: One of the discrete limits in :class:`MaxLoadSB2`.
         :raises ConnectionError: If not connected/negotiated to the device.
         :raises ValueError: If ``load`` is ``MaxLoadSB2.UNKNOWN``.
@@ -371,11 +341,6 @@ class Solarbank2Common(SolixBLEDevice):
         await self._send_command(
             cmd=bytes.fromhex(CMD_SB2_SET_MAX_LOAD), payload=payload
         )
-
-    ##############################
-    #        END UNTESTED        #
-    #          FUNCTIONs         #
-    ##############################
 
     # Telemetry property accessors
 
--- a/SolixBLE/devices/solarbank2.py
+++ b/SolixBLE/devices/solarbank2.py
@@ -170,12 +170,7 @@ class Solarbank2Common(SolixBLEDevice):
     def _build_set_schedule_payload(power_w: int) -> bytes:
         """Build the plaintext payload for cmd 0x405e (set schedule).
 
-        .. warning::
-            There may be a legal & safe power limit below the limit enforced by this software
-            depending on your country of operation.
-            Make sure to comply with all local regulation!
-
-        Produces a uniform 7-day schedule (Mon-Sun, all identical) with the
+        Produces a uniform 7-fday schedule (Mon-Sun, all identical) with the
         same time range (00:00-24:00) and the requested output power.
 
         The caller's session timestamp is appended automatically by
@@ -241,6 +236,11 @@ class Solarbank2Common(SolixBLEDevice):
 
         Sends cmd 405e with a payload that configures every day of the week
         identically: output `power_w` Watts from 00:00 to 24:00.
+
+        .. warning::
+            There may be a legal & safe power limit below the limit enforced by this software
+            depending on your country of operation.
+            Make sure to comply with all local regulation!
 
         :param power_w: Output wattage (0 = charge-only).
         :raises ConnectionError: If not connected/negotiated to the device.
@@ -342,6 +342,11 @@ class Solarbank2Common(SolixBLEDevice):
 
     async def set_max_load(self, load: MaxLoadSB2) -> None:
         """Set the AC output power limit (max load) in watts.
+
+        .. warning::
+            There may be a legal & safe power limit below the limit enforced by this software
+            depending on your country of operation.
+            Make sure to comply with all local regulation!
 
         :param load: One of the discrete limits in :class:`MaxLoadSB2`.
         :raises ConnectionError: If not connected/negotiated to the device.
--- a/SolixBLE/devices/solarbank2.py
+++ b/SolixBLE/devices/solarbank2.py
@@ -237,10 +237,16 @@ class Solarbank2Common(SolixBLEDevice):
         Sends cmd 405e with a payload that configures every day of the week
         identically: output `power_w` Watts from 00:00 to 24:00.
 
-        .. warning::
+        .. attention::
             There may be a legal & safe power limit below the limit enforced by this software
             depending on your country of operation.
             Make sure to comply with all local regulation!
+
+        .. warning::
+            This function writes a full weeks schedule to persistant memory.
+            Excessive use of this function (e.g. in short intervals of 2s)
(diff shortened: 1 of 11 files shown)


SolixBLE/devices/solarbank2.py
```diff
@@ -38,24 +77,308 @@ class MaxLoadSB2(Enum):
...
+        payload = self._build_set_schedule_payload(power_w)
+        await self._send_command(
+            cmd=bytes.fromhex(CMD_SB2_SET_SCHEDULE), payload=payload
+        )
+
+    ##############################
+    #          WARNING           #
+    #   Functions below are      #
+    # NOT E2E Tested on Device!! #
+    ##############################
```
flip-dots: Maybe lets avoid adding untested stuff for now, this PR and file is already very large and we can always add it later, ideally using the same CMD_xxx constant style used for other commands.

SolixBLE/devices/solarbank2.py
```diff
@@ -445,3 +768,289 @@ def battery_heating(self) -> bool | None:
...
+    .. note::
+        A pristine, never-paired SB2 may still require a one-time pairing
+        through the Anker app before any BLE client (including SolixBLE)
+        can connect.
+    """
+
+    async def _initiate_negotiations(self) -> None:
+        """Start the legacy base SolixBLEDevice handshake."""
+        _LOGGER.info("SB2: starting legacy base handshake (00xx/08xx, AES-CBC)")
+        await super()._initiate_negotiations()
```
flip-dots, line 778: I think it makes sense to cut this since _initiate_negotiations() already produces a log entry when called.

SolixBLE/devices/solarbank2.py
```diff
@@ -445,3 +768,289 @@ def battery_heating(self) -> bool | None:
...
+                    "it does not require an Anker user ID."
+                )
+            anker_user_id = env_val
+        if isinstance(anker_user_id, str):
+            anker_user_id = anker_user_id.encode("ascii")
+        self._anker_user_id: bytes = anker_user_id
+
+    # Encryption helpers
+
+    def _encrypt_with_static_key(self, plaintext: bytes) -> bytes:
```
flip-dots, line 856: When I originally implemented the prime protocol I think I just stored the cipher text in constants instead of storing the plain text and then encrypting it before sending it since they were constants, but since you now need that functionality I think it would make more sense to implement it in the [_encrypt_payload()](https://github.com/flip-dots/SolixBLE/blob/18849c92ff10125fc9291379603ad4b67eeab788/SolixBLE/prime_device.py#L109) function of prime_device.py. 

In [_decrypt_payload()](https://github.com/flip-dots/SolixBLE/blob/18849c92ff10125fc9291379603ad4b67eeab788/SolixBLE/prime_device.py#L124) it automatically determines if it should use the static or dynamic key depending upon if the dynamic key is set, you should be able to use that to implement something similar for _encrypt_payload()

SolixBLE/devices/solarbank2.py
```diff
@@ -1,24 +1,63 @@
 """Solarbank 2 power station model.
 
+Two variants live in this module:
+
+* :class:`Solarbank2` - uses the legacy base ``SolixBLEDevice`` handshake
```
flip-dots, line 5: I am almost tempted to say that the parts of Solarbank2Prime that might not be specific to the Solarbank2 should be in a separate file, like how I have device.py and device_prime.py so it would be idk like device_solarbank.py, but it seems a tad premature to give it its own class without knowing if its all specific to the Solarbank2 or if its shared with other Solarbank models.

I think its fine to leave for now but it might make sense to move it later, especially with how large this file has gotten.

SolixBLE/devices/solarbank2.py
```diff
@@ -38,24 +77,308 @@ class MaxLoadSB2(Enum):
...
+    The concrete subclasses :class:`Solarbank2` (legacy CBC) and
+    :class:`Solarbank2Prime` (Prime-style GCM) pick the handshake/crypto
+    behavior by their second base class.
     """
 
     _EXPECTED_TELEMETRY_LENGTH: int = 253
 
+    # Post-handshake command path
+
+    async def _send_command(self, cmd: bytes, payload: bytes) -> None:
```
flip-dots, line 136: This is because the timestamp SolixBLE generates always starts at the same time isn’t it and the timestamp for Solarbanks needs to be correct for schedules to work properly right?

Long term I think it makes sense to upgrade [_send_command()](https://github.com/flip-dots/SolixBLE/blob/18849c92ff10125fc9291379603ad4b67eeab788/SolixBLE/prime_device.py#L502) and the negotiation in prime_device.py so that the timestamp is in line with the actual time but I originally avoided doing that so I could keep using constants for the negotiation, it made testing easier since it would always produce the same value, and no supported devices needed accurate time, but I guess that’s not true any more. 

Keeping this here is fine for now but is something I should probably fix eventually and to do that it would be good to have some tests for the negotiation of the solarbank to make sure it still works when I get around to refactoring it. There are some negotiation tests [here](https://github.com/flip-dots/SolixBLE/blob/18849c92ff10125fc9291379603ad4b67eeab788/tests/test_prime.py#L49) and [here](https://github.com/flip-dots/SolixBLE/blob/18849c92ff10125fc9291379603ad4b67eeab788/tests/test_devices.py#L758) which should help.

SolixBLE/devices/solarbank2.py
```diff
@@ -445,3 +768,289 @@ def battery_heating(self) -> bool | None:
...
+        """Handle SB2 multi-fragment telemetry (cmd c405 + c840, pattern 03010f).
+
+        PrimeDevice's variant assumes single-packet telemetry, but SB2 splits
+        large telemetry across multiple ``03010f`` packets with a per-packet
+        ``<index/total>`` nibble byte after CMD. Delegate to the base
+        SolixBLEDevice implementation, which already implements reassembly.
+        After reassembly it calls ``self._decrypt_payload``, which on this
+        class is PrimeDevice's AES-GCM variant.
+        """
+        return await SolixBLEDevice._process_telemetry_packet(self, payload, cmd)
```
flip-dots, line 1051: I think it might make sense to merge the _process_telemetry_packet of prime_device.py and device.py at some point in the future but this is fine for now as long as we have some tests for telemetry packet processing. 

Looking at [test_telemetry_packet_processing()](https://github.com/flip-dots/SolixBLE/blob/18849c92ff10125fc9291379603ad4b67eeab788/tests/test_devices.py#L994) I can see that there are tests for telemetry packet processing of Prime devices and some Solix power stations but none for the Solarbank2, could you add some please.

tests/test_devices.py
```diff
@@ -1136,3 +1143,105 @@ async def test_bad_values(
...
+
+
+# Solarbank 2 set-schedule payload (0x405e) and Solarbank2Prime construction
+# Synthetic user IDs - never the real Anker cloud userId from any capture.
+_FAKE_SB2_USER_ID_STR = "0" * 40
+_FAKE_SB2_USER_ID_BYTES = b"f" * 40
+_FAKE_SB2_USER_ID_ENV = "1" * 40
+
+
+def test_sb2_build_set_schedule_payload_matches_expected_layout():
```
flip-dots: It would be good if all of these tests were [parameterised](https://docs.pytest.org/en/stable/example/parametrize.html) like the other ones so they could test more values. I don't mind if you make up the data for them rather than using captures, the purpose of most of the tests in this module are more about making sure that changes to the code don’t result in unintended changes to the output rather than strictly adhering to observed behaviour.

tests/test_devices.py
```diff
@@ -1136,3 +1143,105 @@ async def test_bad_values(
...
+    expected_hex = (
+        "a10121"
+        "a2020101"
+        + "".join(day_block(0xa3 + 4 * d) for d in range(7))
+        + "fd050300112233"
+    )
+    assert result.hex() == expected_hex
+
+
+@pytest.mark.parametrize("bad_power", [-1, 801, 1000])
```
flip-dots: Its more of a nitpick than anything but I like to use pytest.param(a, b, ..., n, id="useful_test_id") in all parameterised tests so that if a test fails there is an ID which gives me an initial indication of what the issue might be. That being said I am not super fussed about this since its kind of obvious.

So for these it might be something like "negative", "one_over", and "super_massive".

tests/test_devices.py
```diff
@@ -1136,3 +1143,105 @@ async def test_bad_values(
...
+def test_sb2_build_set_schedule_payload_rejects_out_of_range(bad_power):
+    """Out-of-range wattage raises ValueError."""
+    with pytest.raises(ValueError):
+        Solarbank2Common._build_set_schedule_payload(bad_power)
+
+
+def test_sb2_build_set_schedule_payload_uses_fresh_nonce_per_call():
```
flip-dots: You know with how many damn devices SolixBLE now supports I think I might have to reconsider how tests are structured since this file is kind of massive. This is fine for now though.

scripts/frida_sb2_dart.js
```diff
@@ -0,0 +1,548 @@
...
+ * by Harvey Lelliott (@flip-dots), MIT-licensed.
+ * 
+ * Known issue: plaintext parsing is not implemented. Use the captured
+ * session key + IV to decrypt the encrypted payloads externally.
+*/
+
+// Utilities
+
+function log(msg) {
+    console.log(msg);
```
flip-dots: Hmmm. 

Since its already here I guess you it makes sense to use the log function from [frida_2.js](https://github.com/flip-dots/SolixBLE/blob/18849c92ff10125fc9291379603ad4b67eeab788/scripts/frida_2.js#L13-L27) instead of removing it.

docs/source/app_decoding.rst
```diff
@@ -230,6 +230,35 @@ Below are example outputs of the patched app with the Frida script.
...
+When Java Cipher hooks aren't enough
+------------------------------------
+
+The hooks in ``frida.js`` intercept Java-side ``javax.crypto.Cipher``
+calls, which is sufficient for devices whose Anker-app integration
+performs BLE crypto in Java. Other devices (notably the
+:doc:`Solarbank 2 <solarbank2>`) instead perform their per-session BLE
+AES inside the **Dart AOT** code in ``libapp.so`` using the
+``pointycastle`` library. The Java Cipher hooks never fire for that
+traffic, so the session key cannot be recovered through them.
```
flip-dots, line 242: I am impressed with the lengths you have gone to update the docs, have you tested that this all renders properly? I forget the exact commands to build the docs locally but I think its something like `cd docs` and then `make html` and then in the output folder there is an `index.html` which you can open in a browser which should contain the local changes. You might need to install the stuff in docs/requirements.txt first.

SolixBLE/devices/solarbank2.py
```diff
@@ -38,24 +77,308 @@ class MaxLoadSB2(Enum):
...
+        #   bytes [2:4]  end_min   u16 LE
+        #   bytes [4:6]  power_W   u16 LE
+        #   bytes [6:8]  unknown constant, always 0x0050 LE in captures
+        # 00:00-24:00 -> start=0, end=1440.
+        sched_struct = (
+            (0).to_bytes(2, "little")
+            + (1440).to_bytes(2, "little")
+            + power_w.to_bytes(2, "little")
+            + bytes.fromhex("5000")
+        )
```
flip-dots, line 204: I think I remember seeing a better way of packing bytes using something like [construct](https://github.com/construct/construct) but I am not familiar with it and that kind of structural change is probably better to implement later, I think its fine as long as there are enough tests to make sure I don’t break it when refactoring, so like 3+ test calls of this with different parameters.

SolixBLE/devices/solarbank2.py
flip-dots: That being said I don't want to waste the work already here that might be useful for someone who wants to test/refine this stuff in future but I am not sure where else to put it. Maybe we add an experimental section to the Solarbank2 docs with this layout info in it?

SolixBLE/devices/solarbank2.py
```diff
@@ -445,3 +768,289 @@ def battery_heating(self) -> bool | None:
...
+    traffic. Does not require an Anker cloud user-id.
+
+    This is the recommended default for end users - no cloud-side
+    user-id capture needed. Use :class:`Solarbank2Prime` only if you need
+    to mimic the Anker app's exact handshake (e.g. for protocol research).
+
+    .. note::
+        A pristine, never-paired SB2 may still require a one-time pairing
+        through the Anker app before any BLE client (including SolixBLE)
+        can connect.
```
flip-dots, line 772: Apologies if this is already explained elsewhere and I just missed it, but how would one go about getting this cloud ID, does the app show it anywhere?, if you use the legacy protocol is it included in the telemetry data? or might there be some command in the legacy protocol to request it that could be found in the dumps?

SolixBLE/devices/solarbank2.py
```diff
@@ -445,3 +768,289 @@ def battery_heating(self) -> bool | None:
...
+        await super()._initiate_negotiations()
+
+
+class Solarbank2Prime(Solarbank2Common, PrimeDevice):
+    """
+    SolarBank 2 power station with the Anker-Prime-style handshake.
+
+    Uses 40xx/48xx negotiation across 8 stages and AES-GCM for session
+    traffic. Requires an Anker user-id. SB2 firmware whitelists user-ids
+    and rejects unknown values with RX 4827 = ``09 a1 02 b4 00``.
```
flip-dots, line 787: It wouldn’t happen to accept a 0000000 user ID would it? Its probably wishful thinking but we have seen Anker take a few [shortcuts](https://github.com/flip-dots/SolixBLE/issues/22#issuecomment-4543481408) when it comes to security so I wouldn’t be completely surprised.

SolixBLE/devices/solarbank2.py
```diff
@@ -445,3 +768,289 @@ def battery_heating(self) -> bool | None:
...
+
+    Differences from Anker Prime power stations:
+
+    * Per-session random ECDH private key (Prime uses a hardcoded one).
+    * Different stage-0 through stage-4 plaintexts (live timestamps, extra
+      TLVs).
+    * Stage-6 sets the timezone (TX 4022); stage-7 re-sends the user-id
+      (TX 4027) session-encrypted.
+    * Post-stage-5 payloads use the ``fe 05 03 <ts>`` trailer rather than
```
flip-dots, line 796: Its interesting that it uses the timestamp format that Solix devices seem to use from [device.py](https://github.com/flip-dots/SolixBLE/blob/18849c92ff10125fc9291379603ad4b67eeab788/SolixBLE/device.py#L801) rather than the timestamp format used in [Prime devices](https://github.com/flip-dots/SolixBLE/blob/18849c92ff10125fc9291379603ad4b67eeab788/SolixBLE/prime_device.py#L521), it would be nice if Anker would make up their damn mind.

I suppose its probably something I should also look into to see if I can make the existing code more portable but its fine for now.

SolixBLE/devices/solarbank2.py
```diff
@@ -445,3 +768,289 @@ def battery_heating(self) -> bool | None:
...
+    """
+    SolarBank 2 power station with the Anker-Prime-style handshake.
+
+    Uses 40xx/48xx negotiation across 8 stages and AES-GCM for session
+    traffic. Requires an Anker user-id. SB2 firmware whitelists user-ids
+    and rejects unknown values with RX 4827 = ``09 a1 02 b4 00``.
+
+    Differences from Anker Prime power stations:
+
+    * Per-session random ECDH private key (Prime uses a hardcoded one).
```
flip-dots: I suppose I should probably consider making the other implementations use dynamic ones as well. I did it that way because it made testing easier and I am not super concerned about this type of security which would require close proximity to exploit, but I suppose I should fix it... eventually...

SolixBLE/devices/solarbank2.py
```diff
@@ -445,3 +768,289 @@ def battery_heating(self) -> bool | None:
...
+            if not env_val:
+                raise ValueError(
+                    "Solarbank2Prime requires the Anker cloud-side userId. "
+                    "Either pass it to the constructor "
+                    f"or set the {ENV_SB2_ANKER_USER_ID} environment variable. "
+                    "Alternatively, use the Solarbank2 (legacy CBC) class - "
+                    "it does not require an Anker user ID."
+                )
+            anker_user_id = env_val
+        if isinstance(anker_user_id, str):
```
flip-dots, line 842: There should probably be a length check here if people need to be manually entering Anker account IDs, maybe even some basic regex to make sure its sane since I would not be surprised if people tried to use their emails in it. Some tests as well I guess.

flip-dots (commented):
You know I gotta hand it to you, this is some great work, its highlighted quite a few shortcomings with the existing abstractions I made which I should probably fix at some point, hell most of my review comments are just me highlighting things for future me to do. 

In theory my availability should be a lot better now that my university exams are sorted so feel free to reach out if you need additional clarification or some help with this.

SolixBLE/devices/solarbank2.py
ConniShiva: Hi,
i've been running julians PR-branch for controlling my Solarbank 2 E1600 AC, as I wanted to control two options only available through the APP and can vouch for these functions at least on my E1600 AC (which is different in hw setup from the PRO) but the captures looked the same for these on my device so i tried them.

I have two additional controls added in my local branch, which are very specific to the AC ( "enable  1000w emergency power through AC socket without grid" and "charge battery from grid for X minutes" ) , which I would like to add when this PR is done

jul1an-s: Either way is fine for me, now that we have confirmation that it works I'd simply remove that tag. Thanks for confirming @ConniShiva

SolixBLE/devices/solarbank2.py
jul1an-s: I had exactly the same thoughts and couldn't decide either :D The solarbank family is quite big nowadays

SolixBLE/devices/solarbank2.py
jul1an-s: If I recall correctly the timestamps should at least be higher than the negotiation timestamp. It's not really comparing anything to an actual realtime clock if I recall it correctly. But it's been some time and I'm not sure anymore.

Yes I think so too, but I the PR was super big already and I didn't want to refactor everything with only limited testing capabilities

tests/test_devices.py
jul1an-s: Yeah if this keeps growing this would be massive 😄

docs/source/app_decoding.rst
jul1an-s: No worries, it's not done without the documentation ;-) 
Actually I consider this part of figuring out the additional crypto libs my main contribution. The rest is just following/extending your templates with the new device knowledge.

Yes I have run the build and quickly skimmed through it. Didn't notice any obvious flaws. 

I think your docu build pipeline in HA-Solix-Ble can be used to view the docs build of this PR if I recall it correctly

**Edit**: It's even in the auto build pipeline in this PR :-) https://solixble--36.org.readthedocs.build/en/36/

Ah dammit, looking at it I just discovered I put the wrong script name into the docs :-/

SolixBLE/devices/solarbank2.py
jul1an-s: So the ideal workflow for the future is:

- We reverse engineer the pairing process (not the handshake, the actual first time pairing with the button press).
- I am 99% sure this id is transferred during pairing
- We implement our own pairing in SolixBLE
- User can set whatever ID he/she pleases

In the meantime maybe it's possible to get the ID from the cloud API, but I haven't tried because I knew mine from the captures

SolixBLE/devices/solarbank2.py
jul1an-s: No it rejects a random user ID. But as I said above, reverse engineering the pairing process is probably very easy now. I simply ran out of time

SolixBLE/devices/solarbank2.py
jul1an-s: Hm true that's interesting... Good catch!

SolixBLE/devices/solarbank2.py
jul1an-s: Yeah I'm pretty much with you here. (And Anker as well)
Whoever can hack this can probably also just cut the wire. 

Claude put that nit-picky comment there and I didn't remove it. I didn't want to call you out or whatsoever. Sorry. I'll remove it.

jul1an-s:
Thank you so much for taking your time to review this big blob so thoroughly. I appreciate it a lot!

I will consider every one of your comments, I just replied to the obvious ones where I don't need to change much on my side. 

With that being said my current availability is very poor for the next few months, so I don't know when I'll find the time to work on this again. If someone else wants to pick this up in the meantime I'm fine with it and I'm happy to help.

jul1an-s:
Noticed a bug today in "production": 

Device reports input_cutoff_data = 100 %

**Possible causes**:
- Maybe due to latest firmware update of the device, which now uses input_cutoff_data properly (seemed to be always  5 or 10 in the past) 
- I triggered this behavior by changing the output_cutoff_data using my [set_reserved_power()](https://github.com/jul1an-s/SolixBLE/blob/0b9b1073c056f6c7042a3116141c0be4f262ddbe/SolixBLE/devices/solarbank2.py#L318) function.

Either way this crashes the HA integration. Needs fix before merge @jul1an-s .

<details>
<summary>Error message in HA</summary>

```
Logger: homeassistant.components.sensor
Source: helpers/entity_platform.py:455
integration: Sensor (documentation, issues)
First occurred: 09:02:41 (1 occurrence)
Last logged: 09:02:41

Error while setting up solix_ble platform for sensor: 100 is not a valid SBPowerCutoff
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 455, in _async_setup_platform
    await asyncio.shield(awaitable)
  File "/config/custom_components/solix_ble/sensor.py", line 200, in async_setup_entry
    SolixSensorEntity(
    ~~~~~~~~~~~~~~~~~^
        device,
        ^^^^^^^
    ...<5 lines>...
        state_class=None,
        ^^^^^^^^^^^^^^^^^
    )
    ^
  File "/config/custom_components/solix_ble/sensor.py", line 1117, in __init__
    self._update_updatable_attributes()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/config/custom_components/solix_ble/sensor.py", line 1131, in _update_updatable_attributes
    attribute_value = getattr(self._device, self._attribute_name)
  File "/usr/local/lib/python3.14/site-packages/SolixBLE/devices/solarbank2.py", line 695, in input_cutoff_data
    return SBPowerCutoff(self._parse_int("b6", begin=1))
  File "/usr/local/lib/python3.14/enum.py", line 707, in __call__
    return cls.__new__(cls, value)
           ~~~~~~~~~~~^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/enum.py", line 1192, in __new__
    raise ve_exc
ValueError: 100 is not a valid SBPowerCutoff
```
</details>

thomluther:
The Input cutoff field is now used as max soc, but only if the device firmware supports it.

jul1an-s:
Hi @thomluther,

thank you for sharing. Could you elaborate a little further? The behavior of the field is not clear to me: 

1. Are you talking about telemetry field b6 or TX field a4 of CMD 4067 or both?
2. b6 was 10% at the beginnning. I submitted CMD 4067 with a4 = 5% During the next hours I received b6 = 5%. The next morning I received b6 = 100 and after a HA reboot I received b6 =5% again. Do you have explanation for this behavior?

thomluther:
> 1. Are you talking about telemetry field b6 or TX field a4 of CMD 4067 or both?

This is the old SOC reserve command structure (4067 or MQTT 0067):
```
    "a2": {
        NAME: "set_output_cutoff_data",  # 10 | 5 %
        TYPE: DeviceHexDataTypes.ui.value,
        STATE_NAME: "output_cutoff_data",
        VALUE_OPTIONS: [5, 10],
    },
    "a3": {
        NAME: "set_lowpower_input_data",  # 5 | 4 %
        TYPE: DeviceHexDataTypes.ui.value,
        STATE_NAME: "lowpower_input_data",
        VALUE_FOLLOWS: "set_output_cutoff_data",
        VALUE_OPTIONS: {5: 4, 10: 5},
    },
    "a4": {
        NAME: "set_input_cutoff_data",  # 10 | 5 %
        TYPE: DeviceHexDataTypes.ui.value,
        STATE_NAME: "input_cutoff_data",
        VALUE_FOLLOWS: "set_output_cutoff_data",
        VALUE_OPTIONS: {5: 5, 10: 10},
    },
```
Those 3 fields had only 2 options each, and their state was in 0405 messages:
```
            "b4": {NAME: "output_cutoff_data"},
            "b5": {NAME: "lowpower_input_data"},
            "b6": {NAME: "input_cutoff_data"},
```

I'm not sure what that field was used for, it was inherited from Solarbank 1. Maybe it was always ignored.

The new SOC limit command structure uses different fields except a2 (4067 or MQTT 0067):
```
    "a2": {
        NAME: "set_min_soc",
        TYPE: DeviceHexDataTypes.ui.value,
        STATE_NAME: "min_soc",
        VALUE_MIN: 5,  # 5 % for SB2, 1 %  for SB3
        VALUE_MAX: 20,
        VALUE_STATE: "min_soc",
    },
    "a5": {
        NAME: "set_max_soc",
        TYPE: DeviceHexDataTypes.ui.value,
        STATE_NAME: "max_soc",
        VALUE_MIN: 80,
        VALUE_MAX: 100,
        VALUE_STATE: "max_soc",
    },
    "a6": {
        NAME: "set_backup_soc",
        TYPE: DeviceHexDataTypes.ui.value,
        STATE_NAME: "backup_soc",
        VALUE_MIN: 0,
        VALUE_MAX: 99,
        # ensure backup is soc_min < backup < soc_max
    },
    "a7": {
        NAME: "set_backup_soc_switch",
        TYPE: DeviceHexDataTypes.ui.value,
        STATE_NAME: "backup_soc_switch",
        VALUE_OPTIONS: {"off": 0, "on": 1},
        VALUE_STATE: "backup_soc_switch",
    },
```

The output_cutoff_data b4 field was always the SOC reserve, and is now also used as new min SOC value. The b6 field is presenting the used charge SOC limit, that is typically identical with the max SOC. The eb field is new and shows the max SOC setting from the command:
```
    "b4": {NAME: "min_soc"},
    "b5": {NAME: "lowpower_input_data"},
    "b6": {NAME: "active_charge_soc"},
    "eb": {NAME: "max_soc"},
```

It could also be that the original field always had the same meaning 'input_cutoff_data' => limit for the charge, but maybe this is ignored by the firmware, if not larger than min soc, or 'output_cutoff_data', this the battery was always charged to 100%, since the old SOC reserve command always set this field the same value as the minimum...

Anyway, you need to find a way to recognize from the installed firmware level or 'eb' field value, whether the old or the new SOC command structure should be used. The old command should not be used anymore with the newer firmware, since it may invalidate the max soc setting.

I should also note that Solarbank 2 does not support the backup soc or switch settings to my knowledge, but the App always sends all 4 fields, where the 2 backup fields are always 0. Not sure if they are really needed in the command structure.

SolixBLE/devices/solarbank2.py
jul1an-s: This log output is meant to let the reader now whether legacy or prime style handshake was used for the connection. _initiate_negotiations() does not provide this information. Is it ok to leave it in there for now?

SolixBLE/devices/solarbank2.py
jul1an-s: Does this still hold true with all the new developments in the other devices? If yes, I'll move it. Please let me know.

SolixBLE/devices/solarbank2.py
jul1an-s: Done, all tested and working https://github.com/flip-dots/SolixBLE/pull/36/commits/9eec57ba465486cf01158d3586096ee1589c5942

docs/source/app_decoding.rst
jul1an-s: Fixed https://github.com/flip-dots/SolixBLE/pull/36/commits/2fe530dcb5bdd3d71852098be63fe1ef2309dfe6

SolixBLE/devices/solarbank2.py
jul1an-s: The problem is that I don't know any pattern of possible user_ids. I know mine, and I'll add a length check for it as you requested. But this could also cause problems if they are not fixed length ids. 

I'd need more samples for a sensible regex check.

tests/test_devices.py
jul1an-s: Done https://github.com/flip-dots/SolixBLE/pull/36/commits/208bc891428e4f0ea1f0a54b7644b5562abcea6d

tests/test_devices.py
jul1an-s: Done https://github.com/flip-dots/SolixBLE/pull/36/commits/208bc891428e4f0ea1f0a54b7644b5562abcea6d

SolixBLE/devices/solarbank2.py
jul1an-s: Resolved https://github.com/flip-dots/SolixBLE/pull/36/commits/add6b133bc8ce91248b9171c026da68581c92b31

SolixBLE/devices/solarbank2.py
jul1an-s: I guess you are referring to the original SB1 PR:
https://github.com/flip-dots/SolixBLE/pull/27/changes#diff-bb11d3f89e45b017ddaad80169f83db36e4d169aa0b45bc23d64489d2620d4cfR408

This uses struct. This PR has not been merged yet so I stuck with the existing SolixBLE style of doing this.

However I agree that a harmonized schedule build across solarbanks would be good. If you want it in SB1 style using structs let me know.

**The tests have been added.**

scripts/frida_sb2_dart.js
jul1an-s: Done, my bad, thanks for pointing it out. https://github.com/flip-dots/SolixBLE/pull/36/commits/bb4dcc6244117d399bc9fffd46275f3671a7f1e9

jul1an-s:
@flip-dots I think we're ready to merge. No need to prolong this any longer. 

I will tackle the input cutoff problem later when I have time and I'm planning to rework the set power to make it clear we are writing a schedule there. 
But this will be another PR.

SolixBLE/devices/solarbank2.py
flip-dots: Sure, but maybe don't call the old style "legacy" just yet given that the new one is very experimental. I think its only fair to call it legacy when the newer style is able to fully supersede it (including being able to generate these IDs).

SolixBLE/devices/solarbank2.py
flip-dots: Fair enough. Given that this needs someone to go through the effort of finding the ID I think this needs to remain experimental for now. I am fine with merging it but it needs to be clearly labelled in the docs as experimental.

SolixBLE/devices/solarbank2.py
flip-dots: It holds true for prime devices, the older style Solix devices do not use encryption with a static key for the first few messages, those are all in plain text.

SolixBLE/devices/solarbank2.py
```diff
@@ -445,3 +754,298 @@ def battery_heating(self) -> bool | None:
...
+                    f"SB2 ECDH shared secret derived: {self._shared_secret.hex()}"
+                )
+
+                # TX 4022 - set timezone. Plaintext layout:
+                #   a1 04 <session-ts>  a2 00  a3 04 <tz_offset_seconds_LE>
+                #   a5 <len> <POSIX TZ string>
+                # The tz_offset is signed LE seconds; CEST in capture was -7200
+                # (= -2h). We hardcode that for now - proper localtime detection
+                # is a TODO.
+                tz_str = b"CET-1CEST,M3.5.0,M10.5.0/3"
```
flip-dots, line 994: I think this is too important to not implement. #45 changes how negotiations work so the time is correct for other devices where that matters (e.g chargers/power banks with the built in clock) and there is some [code](https://github.com/flip-dots/SolixBLE/pull/45/changes#diff-428387c407028a3f46026bb9c6e9a434041d6e49b7662ced1362e922b08bb5ccR127) in that for generating time zones that might be useful.

You could also wait for that code (or something similar which fixes times) to be merged since its a relatively high priority and I will probably end up addressing it soon.

SolixBLE/devices/solarbank2.py
```diff
@@ -445,3 +754,298 @@ def battery_heating(self) -> bool | None:
...
+                tz_offset = (-7200).to_bytes(4, "little", signed=True)
+                pt = (
```
flip-dots, line 996: Does the timestamp need to be at the beginning of the packet rather than the end?, usually its at the end. If it works with it at the end you can probably use _send_command() to have it auto-generated instead of needing to generate it for all of these.

SolixBLE/devices/solarbank2.py
```diff
@@ -38,24 +77,294 @@ class MaxLoadSB2(Enum):
...
+        Plaintext layout (without the ``fe 05 03 <ts>`` trailer added by
+        ``_send_command``)::
+
+            a1 01 21              command marker
+            a2 02 01 00           constant 0x00 flag (purpose unknown)
+            a3 02 01 <state>      0 = ON, 1 = OFF (the "light_off_switch" bit)
+        """
+        return bytes.fromhex(f"a10121a2020100a30201{0 if light_on else 1:02x}")
+
+    async def set_light_switch(self, light_on: bool) -> None:
```
flip-dots, line 273: For these kinds of things the rest of the library uses dedicated turn_off and turn_on functions (e.g [here](https://github.com/flip-dots/SolixBLE/blob/cb684b333467e7b1582509a67770223af549564a/SolixBLE/devices/c800.py#L331-L349)) it would be best to stick to that convention.

flip-dots: That and it makes it easier to add to the Home Assistant integration since it expects switches to work that way.

SolixBLE/devices/solarbank2.py
```diff
@@ -38,24 +77,294 @@ class MaxLoadSB2(Enum):
...
+        :param light_on: ``True`` to turn the light on, ``False`` to turn it off.
+        :raises ConnectionError: If not connected/negotiated to the device.
+        """
+        payload = self._build_set_light_payload(light_on)
+        await self._send_command(
+            cmd=bytes.fromhex(CMD_SB2_SET_LIGHT), payload=payload
+        )
+
+    # Mapping for 3rd field reserved power command
+    _RESERVED_POWER_A3_MAP: dict[int, int] = {5: 4, 10: 5}
```
flip-dots, line 285: Why put it here instead of at the top of the file with the rest of the constants?

SolixBLE/devices/solarbank2.py
```diff
@@ -38,24 +77,294 @@ class MaxLoadSB2(Enum):
...
+            cmd=cmd,
+            payload=encrypted,
+        )
+        _LOGGER.debug(f"SB2 _send_command cmd={cmd.hex()} packet={packet.hex()}")
+        await self._client.write_gatt_char(UUID_COMMAND, packet)
+
+    # 0x405e set-schedule
+
+    @staticmethod
+    def _build_set_schedule_payload(power_w: int) -> bytes:
```
flip-dots, line 170: As per [this](https://github.com/flip-dots/SolixBLE/pull/27#issuecomment-5164585656), if you are able to use the FamilyLoadSchedule from #27 that would be really good. If not then it would be good for this code to use the same dataclass structure as is used there. I feel like this is just a tad too complicated to put in the device class and it would also be easier to test and document if it was separate.

SolixBLE/devices/solarbank2.py
```diff
@@ -445,3 +754,298 @@ def battery_heating(self) -> bool | None:
...
+        _LOGGER.debug(f"SB2 stage 0 TX 4001 packet: {packet.hex()}")
+        await self._client.write_gatt_char(UUID_COMMAND, packet)
+
+    async def _process_negotiation(self, cmd: bytes, payload: bytes) -> None:
+        """SB2 handshake state machine (overrides PrimeDevice's variant)."""
+
+        # Decrypt for logging; _decrypt_payload picks static vs session key
+        # automatically based on whether _shared_secret has been set.
+        try:
+            decrypted = self._decrypt_payload(payload)
```
flip-dots, line 909: Is this needed? [_decrypt_payload()](https://github.com/flip-dots/SolixBLE/blob/cb684b333467e7b1582509a67770223af549564a/SolixBLE/prime_device.py#L124) should automatically determine when to use the static/session key based on if the session key variable is None or not.

flip-dots (commented):
There are a few things that I would like to see addressed, most importantly the statically defined time zone, though there is some code which might help you there, that and the bits of code from the solarbank1 that might be compatible which could simplify things a lot. 

Its worth noting that there are a few things in progress at the moment which solve a few of the problems you have had to solve yourself (timestamps, time-zones, negotiation) and you might want to wait for those to get merged before continuing on (though I don't mind if you don't want to wait).

jul1an-s:
Yeah I am aware that there are a lot of things going on in this project right now which make this lib pretty much a moving target.
That's why I wanted to get this merged before the entire core changes and forces me to redo most of this PR.

Honestly I'm not sure if my availability for this project allows me to get this merged given that some people appear to be working at full auto-vibe code speed. 
I want keep testing every change I do E2E with the device and also review every line before release. But this does of course take time and will always be slower than an automated process. Especially with a job and other things to do as well.

I can try to work on your latest comments soon to get this merged before the great refactor. But if I need to rework and retest the entire inner machinery I guess I'll just leave it to someone else with better availability. 
We could also get this merged and have the refactor load on the AI assistants if you plan to accept this kind of contribution for this project. It won't be much overhead on their side.

Please let me know how you want to proceed

flip-dots:
I 100% get it, its really easy to get drawn into something and then find out its a massive hassle with a seemingly infinite number of change requests and a moving target, I experienced the exact same thing when I added support for Bluetooth Hue lights to Home Assistant, first it was the tests, then its the layout/structure, then they deprecated an API I used, and then they added all these requirements for 3rd party libraries to be built with CI/CD and be auditable and it was a total nightmare, I think it ended up taking over a year for all of it to go through and even now there are still issues with it.

There is a good reason for it though, as the main developer of this thing I am taking on responsibility for maintaining it which means I need to be able to understand and maintain it, as well as anyone else who might need to change it and I don't have the hardware to test, so making unilateral changes later to things like packet decoding and negotiation is quite risky, tests can help mitigate most of it but not all. 

In terms of the more practical aspects I am hoping some of these soon to be implemented changes (like timestamps and timezones being handled by the base classes) will be able to significantly shrink this PR which shrinks the amount of code that I have to maintain that I am not able to test on hardware. I don't expect the big PRs that change a lot of the things to be merged soon, there are many issues with them as it stands, I suspect I will end up making my own which implement just the minimum. In terms of timescales I am happy to leave this PR as draft for as long as needed and I might end up building something to implement timestamps and timezones for all devices in the next few days since I now have a device where the time actually matters (one of those desktop chargers which has a clock).

If your availability does not allow for you to sort all of this out yourself I don't mind implementing the improvements myself and I could probably do that some time in the next month or two after the timestamp/time-zone changes are done, so long as you are willing to test it and make sure it works and send me any test data I might need.

jul1an-s:
I totally understand that you as a maintainer have a great interest in keeping the codebase lean and maintainable.

I'm sorry that you had so much hassle with your HA project :-( However I can understand them having thousands of users and hundreds of contributors. A project of that size needs this kind of thoroughness (especially given that they are doing very critical things like door locks etc).
It really depends on where you want to go with this project. Imo this started as a damn cool POC with a hardcoded private key and lot's of device templates which were simple placeholders without any use (no offense, I would have done the same thing and this was exactly the right approach 🙂 ). 
Of course it's up to you to decide and I understand your motives. 

Nevertheless you should consider two aspects: 
1. We don't even know whether the Anker protocol allows this level of generalization of the protocol. From what I've seen, each device seems to do whatever it pleases and a common handshake might never work across all devices. Actually Anker might need an architect like you 😉 Whenever you change the core of this lib, you will need to retest EVERY device that relies on it. I know you heavily rely on automated testing which is good, but it only gets you so far.
2. It's just code. It's not set in stone. You can roll back every PR, you can remove any device implementation you don't like at any time.

Regarding the size of this PR we could simply drop the frida_dart script which is 1/3 of the PR and the "SolarbankPrime" path which is probably another 1/3 and not wired to your HA repo anyways. I checked it in to document my findings for future contributors, but ofc it introduces maintenance overhead for you. It can also rest on my harddrive like other stuff I discovered which is not part of this PR (OTA flashing etc)

Sure we can leave this open if you have time to work on it in the close future, but I really wanted to ship this to "normal" users in HA soon given some comments in the HA repo which suggest this is already being used by some people 🤔 Especially with Anker having officially denied all hopes for an official solution for this device.

MrChurch:
Hi Julius,

we have now integrated an owned Solarbank 2 E1600 AC (A17C0) locally via BLE. Your dynamic/Prime-style handshake work was very helpful, and a few observations from our captures may be useful for your PR:

- The complete observed sequence is:
  `4001 -> 4801 -> 4003 -> 4803 -> 4029 -> 4829 -> 4005 -> 4805 -> 4021 -> 4821 -> 4022 -> 4822 -> 4027 -> 4827 -> 4040`

- `4005/4805` was required before the P-256 ECDH exchange on our device.

- The bootstrap requests `4001`, `4003`, `4029`, and `4005` reused one timestamp. Your implementation already does this; our capture confirms that behaviour.

- The replies to `4022` and `4027` arrived as AES-GCM session frames with pattern `03000f`, not `030001`:
  `RX 4822 (03000f) -> TX 4027 -> RX 4827 (03000f)`.

  Until the session is ready, those two commands need to be routed back into the negotiation state machine. Our first implementation treated `03000f` as normal telemetry, discarded `4822`, and the device disconnected before reaching `4040`.

This may be specific to the A17C0 / Solarbank 2 AC dynamic path, so we would keep it guarded and separate from the legacy SB2 handshake.

Best regards,
Lumi

PR #38, opened 2026-06-21
Add working C1000 Gen 2 telemetry + AC control (fixes #22)

### LINKED ISSUE #22: Support for Solix C1000 Gen 2?

lukasbraach:
Hey all,

This PR makes the C1000 Gen 2 (A1763) actually work: telemetry plus AC and DC on/off control, confirmed on real hardware. The `C1000G2` class was a non-working, telemetry-only stub — it negotiates the encrypted session fine but never receives any telemetry and exposes no control (#22). The Gen 2 uses the same ECDH/AES-CBC framing as the gen-1 models but with different command codes, which this PR wires up.

### Hardware verification

Load-tested the Gen 2 telemetry against a known load (hairdryer on AC, MacBook on USB-C) and matched the official Anker app:

- `power_out` (`a6[1:3]`) and `ac_power_out` (`a7[2:4]`) confirmed.
- AC output status is **`a7[1]`** (latches `00`/`01` with the relay); DC output status is `b2[1]` with `dc_power_out` = `b2[2:4]`; USB-C per-port follows the same `04 <status> <watts LE>` shape (confirmed via USB-C3).
- Note: `a4` byte 22 is **not** the AC state — it stays `01` even with the port physically off — so `ac_output` reads `a7[1]`.

### Base device

- Add an overridable `_TELEMETRY_COMMANDS` class attribute (default the existing `c402`/`4300`/`c405`) so a model can declare its own telemetry command codes.
- Add an overridable `_post_connect()` hook, called after a negotiated connection (and on every auto-reconnect), for device-specific setup such as sending a subscribe command. Default is a no-op, so other devices are unaffected.

### C1000G2

- Subscribe to telemetry on connect (`_post_connect` sends cmd `4100` / `a10121`); without this the Gen 2 streams nothing.
- Route the Gen 2 telemetry commands `c421`/`c900` (`_TELEMETRY_COMMANDS`).
- Add `turn_ac_on()`/`turn_ac_off()` (cmd `4101`) and `turn_dc_on()`/`turn_dc_off()` (cmd `4102`), matching the gen-1 C1000 API.
- Decode `ac_output` (`a7[1]`), `dc_output` (`b2[1]`) and per-port power/status.

### Tests

- Add `C1000G2` value cases from telemetry frames captured on a real C1000 Gen 2 (A1763), covering AC off/on and DC on. The AC pair is identical except for the `a7` param, asserting the `a7[1]` decode (an `a4`-based decode would report a false `OUTPUT` with the port off).

Let me know if you require further information or collaboration to get this PR merged. Thanks!

### DIFF
--- a/SolixBLE/device.py
+++ b/SolixBLE/device.py
@@ -52,6 +52,11 @@ _LOGGER = logging.getLogger(__name__)
 class SolixBLEDevice:
     """Solix BLE device object."""
 
+    #: Command codes (hex) that carry telemetry for this device. Subclasses can
+    #: override this if their model uses different telemetry command codes
+    #: (e.g the C1000 Gen 2 uses ``c421``/``c900`` instead of ``c402``/``c405``).
+    _TELEMETRY_COMMANDS: tuple[str, ...] = ("c402", "4300", "c405")
+
     def __init__(self, ble_device: BLEDevice) -> None:
         """Initialise device object. Does not connect automatically."""
 
@@ -195,6 +200,15 @@ class SolixBLEDevice:
         if self._disconnect_event.is_set():
             self._disconnect_event.clear()
 
+        # Run any device-specific post-connect setup (e.g sending a subscribe
+        # command to start telemetry). This runs on every (re)connection. Errors
+        # are logged but do not abort the connection; the automatic reconnect
+        # task will retry.
+        try:
+            await self._post_connect()
+        except Exception:
+            _LOGGER.exception(f"Error running post-connect setup for '{self.name}'!")
+
         # Start an automatic reconnect task if its not running already
         if self._auto_reconnect_task is None:
             self._auto_reconnect_task = asyncio.create_task(self._auto_reconnect())
@@ -204,6 +218,17 @@ class SolixBLEDevice:
             self._run_state_changed_callbacks()
 
         return True
+
+    async def _post_connect(self) -> None:
+        """Run device-specific setup after a negotiated connection is established.
+
+        Called by :meth:`connect` once the encrypted session has been negotiated
+        (so :meth:`_send_command` may be used) and on every automatic reconnect.
+        The default implementation does nothing; subclasses can override it to,
+        for example, send a subscribe command to start a telemetry stream (see
+        :class:`~SolixBLE.devices.c1000g2.C1000G2`).
+        """
+        return None
 
     async def disconnect(self) -> None:
         """Disconnect from device and reset internal state.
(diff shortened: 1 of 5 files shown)


SolixBLE/devices/c1000g2.py
```diff
@@ -94,13 +157,20 @@ def ac_output(self) -> PortStatus:
         PortStatus.NOT_CONNECTED signifies off.
         PortStatus.OUTPUT signifies on.
 
+        The AC port status is the first byte of the ``a7`` parameter, mirroring the
```
flip-dots: Its probably not a bad idea to include this kind of information but it might be a bit much for someone who just wants to use the API rather than understand how it works since this text is used to auto-generate the docs. I have taken a look at the sphinx docs and it seems to support [collapsible text](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#admonitions-messages-and-warnings) which might work well for this kind of application, maybe give it a go?

SolixBLE/devices/c1000g2.py
```diff
@@ -116,7 +186,9 @@ def solar_port(self) -> PortStatus:
 
     @property
     def solar_power_in(self) -> int:
-        """Solar/DC Power In.
+        """Solar/DC Power In (watts).
+
+        .. note:: Offset inferred, not yet confirmed on hardware (no solar/DC-in capture taken).
```
flip-dots, line 195: I probably need to find a standard way to define if a feature is untested or confirmed working in the docs at some point instead of each instance describing it, but this is fine for now.

SolixBLE/devices/c1000g2.py
```diff
@@ -164,7 +236,7 @@ def usb_port_c3(self) -> PortStatus:
 
     @property
     def usb_c3_power(self) -> int:
-        """USB C3 Power.
+        """USB C3 Power (watts).
```
flip-dots, line 243: 👍

SolixBLE/device.py
```diff
@@ -52,6 +52,11 @@
 class SolixBLEDevice:
     """Solix BLE device object."""
 
+    #: Command codes (hex) that carry telemetry for this device. Subclasses can
+    #: override this if their model uses different telemetry command codes
+    #: (e.g the C1000 Gen 2 uses ``c421``/``c900`` instead of ``c402``/``c405``).
+    _TELEMETRY_COMMANDS: tuple[str, ...] = ("c402", "4300", "c405")
```
flip-dots, line 58: This is a good idea, especially with how much devices seem to vary.

SolixBLE/device.py
```diff
@@ -195,6 +200,15 @@ async def connect(self, max_attempts: int = 3, run_callbacks: bool = True) -> bo
         if self._disconnect_event.is_set():
             self._disconnect_event.clear()
 
+        # Run any device-specific post-connect setup (e.g sending a subscribe
```
flip-dots, line 203: I like this solution.

SolixBLE/device.py
```diff
@@ -205,6 +219,17 @@ async def connect(self, max_attempts: int = 3, run_callbacks: bool = True) -> bo
...
+    async def _post_connect(self) -> None:
+        """Run device-specific setup after a negotiated connection is established.
+
+        Called by :meth:`connect` once the encrypted session has been negotiated
+        (so :meth:`_send_command` may be used) and on every automatic reconnect.
+        The default implementation does nothing; subclasses can override it to,
+        for example, send a subscribe command to start a telemetry stream (see
+        :class:`~SolixBLE.devices.c1000g2.C1000G2`).
+        """
+        return None
```
flip-dots: Probably makes more sense to use pass here instead of explicitly returning None.

flip-dots (commented):
Nice work. I am happy with most of this, I only have two minor suggestions which would be good to have addressed.

SolixBLE/devices/c1000g2.py
lukasbraach: Alright, sounds good! Only catch: Collapsible has been added in Sphinx version 8.2, while this repo currently uses v7.2.5. I can make the edit and bump the version in this PR, but it might take a bit more validation from your side to check if the docs still generate correctly. Just let me know your preferred way of moving forward.

lukasbraach:
@flip-dots Addressed the 2 open points, waiting on your feedback for the Sphinx version bump

HenriWahl:
@lukasbraach I can confirm this works like a charm! 🙏

SolixBLE/devices/c1000g2.py
flip-dots: Bumping the version of Sphinx is fine, whatever the latest is. If it doesn’t work I don’t mind sorting it out later after this is merged.

lukasbraach: @flip-dots Adding the Sphinx upgrade now...

lukasbraach:
@flip-dots Sphinx is updated to version 8.2.3, added the collapsible.

flip-dots:
Ok, happy to merge this now, thanks for working on this. Support for devices which I don't have is entirely dependent upon people like you who are willing to go above and beyond to get them working, even though it's kind of a pain in the ass, so your work is very much appreciated. 

I plan on doing a release with these changes (and a few others) in the next few days and the update to the Home Assistant integration which uses this should be rolled out soon(ish) after that.

PR #40, opened 2026-07-05
Use default service cache value

flip-dots:
When creating a bleak client to connect to a device the use_service_cache value was set to false, overriding the default. This was done a while back when trying to diagnose an issue with missing services when connecting to a device, this PR reverts that change so the default value of true is used.

This may or may not solve the issue found [here](https://github.com/flip-dots/HaSolixBLE/issues/3#issuecomment-4880601952).


flip-dots:
Its unclear if this makes a difference, people have reported success without it and using it did not resolve the issue this was created to solve so ¯\\_(ツ)_/¯.

PR #41, opened 2026-07-06
Add Anker MagGo 3-in-1 Wireless Charger (A25x7) support

radzio:
Adds a device profile for the **Anker MagGo 3-in-1 Wireless Charger** (model **A25x7**, serial prefix `ASHDKXU`) — phone MagSafe pad + Apple Watch puck + earbuds pad.

### How it works
It's part of the same `8c85…` / `0000ff09` family as the Prime chargers and reuses `PrimeDevice`'s ECDH + AES-GCM negotiation and telemetry framing. Two model-specific details:

- **Subscribe to start telemetry.** Like the C1000 Gen 2, this charger streams nothing until it receives a subscribe command — `0x0200` (`4200` with the encryption flag), payload `a10121`. Sent from `_post_connect()`.
- **Telemetry on `4300`.** The report carries the three wireless pads in TLV params `a2`/`a3`/`a4`, each using the Prime charger per-port shape (`04 <status> <volt LE> <current LE> <power LE>`).

Exposes per-pad power (W) + status and a computed total.

### Testing / confidence
Reverse-engineered from BLE captures (nRF/iOS PacketLogger + on-device probing):
- ✅ **ECDH encryption negotiation confirmed live on real hardware** (connects + negotiates a shared secret).
- ✅ **Per-pad power confirmed** against an Apple Watch charging at 2.8 W → `a2[6:8]` little-endian `0x011d` = 285 → **2.85 W**.
- ⚠️ Per-pad voltage/current fields are present at `[2:4]`/`[4:6]` but their scaling wasn't confidently confirmed for wireless pads, so only **power + status** are exposed here (kept conservative, matching the style of the existing charger profiles).

Companion HA integration PR: flip-dots/HaSolixBLE (adds the model to the dropdown + sensors).

https://claude.ai/code/session_014PHXj8zVt63X4FV8YBag7E


SolixBLE/devices/maggo_3in1.py
```diff
@@ -0,0 +1,115 @@
...
+
+#: Command sent after connecting to start the telemetry stream. Like the C1000
+#: Gen 2, this charger streams nothing until it receives this subscribe command.
+CMD_SUBSCRIBE = "4200"
+SUBSCRIBE_PAYLOAD = "a10121"
+
+
+class MagGo3in1(PrimeDevice):
+    """
+    Anker MagGo 3-in-1 Wireless Charger (model A25x7, serial prefix ``ASHDKXU``).
```
flip-dots: Try to stick to the same documentation layout as the [other](https://github.com/flip-dots/SolixBLE/blob/6208a0bc5269f723c4f976d5e3617344b09f3707/SolixBLE/devices/c300dc.py#L19-L26) modules.

SolixBLE/devices/maggo_3in1.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Use this class to connect and monitor the 3-in-1 wireless charging station
+    (phone MagSafe pad + Apple Watch puck + earbuds pad).
+
+    It uses the same ECDH + AES-GCM encryption and telemetry framing as the
```
flip-dots: I have recently been trying using the ::collapsible:: tag to hide technical details unless someone explicitly looks for them, since someone just using the module might not need to know these details, there are a few examples of how to use it dotted around, such as [here](https://github.com/flip-dots/SolixBLE/blob/6208a0bc5269f723c4f976d5e3617344b09f3707/SolixBLE/devices/c1000g2.py#L160-L162), I think it also makes sense to use that here.

SolixBLE/devices/maggo_3in1.py
```diff
@@ -0,0 +1,115 @@
...
+        send it after every (re)connection.
+        """
+        await self._send_command(
+            cmd=bytes.fromhex(CMD_SUBSCRIBE),
+            payload=bytes.fromhex(SUBSCRIBE_PAYLOAD),
+        )
+
+    @property
+    def pad_1(self) -> PortStatus:
+        """Wireless pad 1 status (TLV ``a2``).
```
flip-dots: Technical details like (TLV ``a2``) should be hidden using a ::collapsible:: tag, like I suggested above.

SolixBLE/devices/maggo_3in1.py
```diff
@@ -0,0 +1,115 @@
+"""Anker MagGo 3-in-1 Wireless Charger model.
+
+.. moduleauthor:: reverse-engineered by community contribution
```
flip-dots: This is meant to be a username / email combination, like [here](https://github.com/flip-dots/SolixBLE/blob/6208a0bc5269f723c4f976d5e3617344b09f3707/SolixBLE/devices/c1000g2.py#L3), not an arbitrary string.

flip-dots: I don't mind if you use your GitHub generated email (mine is 42912136+flip-dots@users.noreply.github.com) but there does need to be actual authorship info.

SolixBLE/devices/maggo_3in1.py
```diff
@@ -0,0 +1,115 @@
...
+
+"""
+
+from ..const import DEFAULT_METADATA_FLOAT
+from ..prime_device import PrimeDevice
+from ..states import PortStatus
+
+#: Command sent after connecting to start the telemetry stream. Like the C1000
```
flip-dots: I am probably nitpicking here but in this repo the C1000 Gen 2 is usually referred to as the C1000G2 and its good to be consistent.

SolixBLE/devices/maggo_3in1.py
```diff
@@ -0,0 +1,115 @@
...
+    def pad_3_power(self) -> float:
+        """Wireless pad 3 power (W).
+
+        :returns: Power delivered by pad 3 or default float value.
+        """
+        if self._data is None:
+            return DEFAULT_METADATA_FLOAT
+        return self._parse_int("a4", begin=6, end=8) / 100.0
+
+    @property
```
flip-dots, line 113: Is there not a dedicated value for this in the telemetry?

SolixBLE/devices/maggo_3in1.py
```diff
@@ -0,0 +1,115 @@
+"""Anker MagGo 3-in-1 Wireless Charger model.
+
+.. moduleauthor:: reverse-engineered by community contribution
+
+"""
+
+from ..const import DEFAULT_METADATA_FLOAT
+from ..prime_device import PrimeDevice
+from ..states import PortStatus
```
flip-dots, line 9: I would really like to see some tests added for this using captures from the device. The tests are a bit unorganised at the moment so don't worry about making it pretty but its good to have something so I can refactor stuff without worrying about accidentality breaking it.

Specifically I would like to see at least two tests for the parsing of values from telemetry data where the values are different (e.g one capture where the output of port 1 is 5w and port 2 is 10w and then another where port 1 is not in use and port 2 is 5w). Ideally these tests would flip every state at least once (e.g port 1 and 2 are tested in use, not in use, at one power level, and at another power level) but I am not super strict about that. I already have a [parameterised](https://docs.pytest.org/en/stable/example/parametrize.html) test [here](https://github.com/flip-dots/SolixBLE/blob/6208a0bc5269f723c4f976d5e3617344b09f3707/tests/test_devices.py#L822) which you just need to add the data for.

In terms of bonus tests which would be nice to have but I am not going to require, it would be good to have a [negotiation test](https://github.com/flip-dots/SolixBLE/blob/6208a0bc5269f723c4f976d5e3617344b09f3707/tests/test_devices.py#L953) since all devices respond differently and more captures could be useful for figuring out more about the protocol in future. In addition a test for [telemetry packet processing](https://github.com/flip-dots/SolixBLE/blob/6208a0bc5269f723c4f976d5e3617344b09f3707/tests/test_devices.py#L1218) would be good to have in case your device is different.

To get the data needed to make these tests I usually use the [demo.py](https://github.com/flip-dots/SolixBLE/blob/main/examples/demo.py) script in debug mode to log the output of the different stages which can then be put into the tests.

SolixBLE/devices/maggo_3in1.py
```diff
@@ -0,0 +1,115 @@
+"""Anker MagGo 3-in-1 Wireless Charger model.
```
flip-dots, line 1: This new device also needs to be added to [demo.py](https://github.com/flip-dots/SolixBLE/blob/main/examples/demo.py).

flip-dots (changes requested):
This is really good, pretty much the only major thing it needs is some tests. Please let me know if you need any help with those.

SolixBLE/devices/maggo_3in1.py
radzio: No dedicated total in the 0x0300 report - it only carries the three per-pad TLVs (a2/a3/a4). The remaining fields aren't a total (a5 is a constant `04ffff`, a6 is zeros/`ff`s, fe is zeros). So I've kept it computed as the sum of the pads, but renamed `total_power` -> `power_out` to match the existing "Total Power Out" HA sensor that reads a `power_out` property.

flip-dots:
Looks good now, I forgot to mention that it needs to be added to the docs and supported devices in the README.md so I have sorted that out myself.

In terms of when a new release with this device will be done, I think I will wait a bit to see if #27 can be included in that release first, so probably in a few days.

thomluther:
Hi @radzio 
I would be interested to add A25X7 also to the [Anker Solix Api library](https://github.com/thomluther/anker-solix-api) for MQTT data (and finally also to the cloud based HA integration)
To do so, you would need to install the library and use the mqtt_monitor tool to watch and [decode MQTT messages](https://github.com/thomluther/anker-solix-api/discussions/222) as well as commands as sent by the mobile App.

The structures are pretty much identical to the BLE messages, but the message codes may differ.
As such, you first have to identify the status request command (for an individual device status message) and the realtime trigger command, which are typically supported by each device. They may be similar to the A2345 charger on '0200' and '020b'.

You can also use the export system module to provide a randomized export from your account, including MQTT messages. I would need that to debug and develop the implementation. When running such an export, you can play around in the App to trigger messages and power or setting changes, which will then become visible in the replay of the messages.

Issue #42, opened 2026-07-09
Unknown session frames aren't reassembled — multi-fragment device posts (e.g. c490) decode only their first fragment


kb1ibt:
## Summary

`_process_notification`'s unknown-`cmd` branch decrypts a **single raw BLE packet**, whereas telemetry frames go through `_process_telemetry_packet`, which reassembles fragments first. Any non-telemetry session frame that spans more than one notification is therefore never reassembled — only fragment 1 reaches `_decrypt_payload`, so it can't decode. The C2000 G2 (A1783) posts exactly such a frame — `cmd=c490`, ~372 bytes cleartext — every ~9 minutes, and it's currently lost.

## Background — what `c490` is

While monitoring an Anker SOLIX **C2000 G2 (A1783)** — which connects on the shared **C1000 G2** family profile (`CapturingC1000G2`) — the device posts an unsolicited frame on its own timer, roughly every 540 s. It's not a request/response we drive.

- Pattern `03010f`, `cmd=c490`, **not** in `_TELEMETRY_COMMANDS`, no future registered → falls into the unknown-message `else` branch.
- Decrypts with the **same AES-CBC session key** SolixBLE already negotiates (`_decrypt_payload`).
- **372 bytes** of cleartext once reassembled — readable ASCII inside (`A1763`, `charging_pps_series_c_0005`), and consecutive frames differ in only a handful of isolated byte positions (i.e. it's plaintext, not still-ciphertext — a CBC change would avalanche).

It carries a useful device-state rollup (displayed SoC, pack voltage, cumulative energy, a runtime estimate). The **payload semantics are out of scope for SolixBLE** — this issue is purely about the transport not delivering the frame intact.

## Root cause

In `SolixBLE/device.py::_process_notification`, the session-message branch splits into:

- **telemetry** (`cmd in _TELEMETRY_COMMANDS`) → `_process_telemetry_packet(payload, cmd)`, which reassembles fragments (`payload[0]` high nibble = index, low nibble = total), strips the header byte, then decrypts;
- **unknown** (`else`) → decrypts `payload` directly.

The unknown branch:

1. **never reassembles** — it operates on one BLE packet, so a multi-fragment frame like `c490` only ever presents fragment 1; and
2. **doesn't strip the fragment-header byte** before decrypt (telemetry does, via `payload = payload[1:]`), which is what the `len(payload) % 16 == 15` realign trick is compensating for on the single-packet path.

[...]

PR #45, opened 2026-07-17
feat: live BLE telemetry for Anker Prime chargers (A2345/A91B2) + C2000 G2 + frame reassembly (#42)

kb1ibt:
## Summary

Adds live, cloud-free BLE telemetry for the Anker **Prime** charger family and the **C2000 Gen 2** power station, plus a shared frame-reassembly fix (#42).

Two foundational changes are what make the newer firmware actually stream:

**1. Computed (live) timestamps throughout the handshake — the major refactor.**
The negotiation and session commands previously **replayed timestamps hard-coded from a single captured app session**: `BASE_TIMESTAMP`, the frozen `a104<ts>` baked into each stage constant (`NEGOTIATION_COMMAND_*`), and the `fe04<ts>` / `fe0503<ts>` command trailers. Anker's replay protection (the reason those timestamps exist) rejects a stale one, so on firmware that enforces it the device negotiated fine but never streamed and dropped the link after ~40s. Every frame is now **built at send time with the live clock** — the init, each negotiation stage, the confer, the registration, and every `_send_command` — via `_ts()` / `_live_negotiation_packet` (Prime, AES-GCM) and the equivalent base path (AES-CBC). Nothing is replayed from a frozen constant anymore. This is why the fixtures had to become timestamp-agnostic (a `frozen_time` fixture pins the clock so the on-wire bytes are reproducible).

**2. Owner binding for hardened firmware.**
Even with live timestamps, the hardened Prime firmware (the C1000 G2 "#22" lockdown, now shipping on some A2345 units) **withholds telemetry until the client proves account ownership**: the `4027` registration must bind the account `owner_user_id`. Add `OWNER_USER_ID` (opt-in; `None` keeps the legacy `a224<uuid>` registration for older firmware) — when set, register `4027 = a104<ts> a228<owner_user_id>`, which is accepted (ack `00`) where the UUID form is rejected (ack `09`).

## What's included

**Prime chargers** (`prime_device` / `prime_charger_250w` / `prime_usb_charger`) — the two changes above, plus: send the real `420b` REALTIME_TRIGGER enable block (`a20a…`, was a no-op bare `a10121`) that starts the `4303` per-port stream; drop the stale app-UUID from the `420a` getter; persist the stage-3 identity (serial / fw / MAC); expose serial, total power, and per-port switches.

**A91B2 240W charging station** (`prime_charging_station_240w`) — full support: base/CBC transport layered over the Prime class; `4a00` snapshot + `4303`/`4302` stream remapped onto one port view; AC-outlet switch states.

**C2000 Gen 2** (`c2000g2` / `c1000g2`) — new model; nested self-delimiting telemetry walkers factored into a shared `parsing` module.

**Frame reassembly (#42)** (`device`) — multi-fragment session frames reassembled with an ATT-MTU length gate so single- and multi-packet telemetry both decode, across GCM and CBC.

**Docs** — guide for obtaining `owner_user_id` without the Anker app.

## Testing — validated on real hardware

| device | model | verified |
|---|---|---|
| **C2000 Gen 2** | A1783 | negotiation, telemetry, per-port decode; cross-checked vs cloud |
| **Prime Charging Station 240W** | A91B2 | negotiation, `4a00`+`4303` stream, per-port + AC-switch decode vs cloud |
| **Prime Charger 250W** | A2345 | negotiation, `ca00`+`4303` stream (incl. the hardened `05F0` owner-binding), per-port decode vs cloud |

Full test suite green (timestamp-agnostic negotiation fixtures + new parsing/reassembly coverage).

## Not hardware-tested

`PrimeCharger160w` and `PrimePowerBank20k` inherit the same shared `_post_connect`, so they pick up these changes but I don't have the hardware to re-test them. The **negotiation handshake order is unchanged vs `main`** (verified) — including the live-timestamp switch, which changes frame *contents* but not the stage sequence. The only post-connect deltas they see are the appended `420b` trigger and the UUID-dropped `420a` (registration stays the legacy `a224<uuid>` since `OWNER_USER_ID` is unset). Protocol-compatible, but a confirmation on a 160w would be appreciated — happy to gate `420b` to the 250w path instead if you'd prefer.


kb1ibt:
Much of this work was done using Apple PacketLogger and comparing it against the app_log.log file from the AppData backup of the same mobile device. Tested against the 3 listed device models, and ran pytest MANY times.

kb1ibt:
### Update — c490 summary fixes + test coverage (`f102712`)

Adds coverage for the two paths that had none — the C2000 G2 `c490` protobuf
summary and the Prime charger telemetry — and fixes the bugs that surfaced once
they were actually exercised.

**c490 device-summary (`C1000G2` / `C2000G2`)**
- `walk_protobuf` was never imported in `device.py`, so the summary path raised
  `NameError` on the first real frame. Nothing hit it before, so it stayed latent.
- The walk ran on the **whole** decrypted frame, but the protobuf is the value of
  the outer `a2` field (a `bin` field with a **2-byte** length the flat TLV parser
  can't read). Result: `summary` filled with garbage (`.20`/`.6`) and every `.23.x`
  property silently returned its default. New `_protobuf_body` strips the `a1`/`a2`
  wrapper and **bounds the slice to `a2`'s declared length**, so the walk sees
  exactly the protobuf and never the trailing `a3` string.
- `walk_protobuf` now records message **containers** (and the wire-3/4 group marker)
  instead of dropping them, matching the reference 116-field decode. Verified over a
  288-frame corpus: leaf values unchanged, no field under-count.

**Prime chargers**
- Promoted the `4303`/`ca00` shared telemetry decode from the A91B2 station up to
  `PrimeUsbCharger`, so the A2345 charger inherits it and decodes its `4303` stream
  (ports remapped `a2`–`a7` → `a4`–`a9`) instead of logging it as unknown. The
  station keeps only its CBC/negotiation and AC-switch specifics.

**Tests: 83 → 123**
- `test_summary.py` (new): the full decrypt → unwrap → walk → property path on real
  captured frames, `_protobuf_body` wrapper/bounding, and defaults-before-first-frame.
- `test_parsing.py`: `read_varint` boundaries; `walk_protobuf` across every wire type
  (container recording, group stop, string vs hex leaves); `walk_lv`.
- `test_reassembly.py`: single-vs-fragment classification, the full-length threshold
  (derived from `_FRAME_OVERHEAD`, not hardcoded), 3-fragment runs, interleaved
  per-command buffers.
- `test_devices.py`: a real A1783 `c421` frame's fields, `expansion_present`, and the
  Prime 250w/station `ca00`/`4a00` snapshot + `4303` stream remap.

Note: the flat `_parse_payload` is intentionally left 1-byte-length only — Anker's
2-byte-length disambiguation is a guess-and-validate heuristic (type must be
`str`/`bin` + a validation pass), so the wrapper strip is handled structurally in
`_protobuf_body` rather than by generalizing the parser every device depends on.

kb1ibt:
I realized I have one more commit to add, the changes to the docs.

kb1ibt:
I split this PR into 5 segments PRs #48 , #49 , #50 , #51 , #52 ; so that way each PR is individually less than 1000 lines modified per PR. since you commented in a [comment on #22 ](https://github.com/flip-dots/SolixBLE/issues/22#issuecomment-5023196167) that the ~3000 was a lot for one PR

PR #48, opened 2026-07-20
feat(ble): reassemble multi-fragment frames for every session command (#42)

### LINKED ISSUE #42: Unknown session frames aren't reassembled — multi-fragment device posts (e.g. c490) decode only their first fragment

kb1ibt:
Split out of #45 per review feedback — the first of several smaller, focused PRs.

Reassembles multi-fragment BLE frames for **every** session command before the cipher, so telemetry and unknown session frames share one reassembler instead of only reassembling telemetry. Fixes truncated/dropped multi-packet frames on commands whose payloads exceed the MTU.

- `device.py`: unify fragment reassembly across all session commands — per-command buffers, single-vs-fragment classification, and a full-length threshold derived from `_FRAME_OVERHEAD` (not hardcoded).
- `tests/test_reassembly.py`: classification, the length threshold, 3-fragment runs, and interleaved per-command buffers.

Closes #42.

This is the base cut of the #45 split; the c490 summary decode, Prime negotiation, Prime device support, and docs will follow as separate PRs.

### DIFF
--- a/SolixBLE/device.py
+++ b/SolixBLE/device.py
@@ -57,12 +57,18 @@ class SolixBLEDevice:
     #: (e.g the C1000 Gen 2 uses ``c421``/``c900`` instead of ``c402``/``c405``).
     _TELEMETRY_COMMANDS: tuple[str, ...] = ("c402", "4300", "c405")
 
+    #: Fixed ff09-frame overhead between the on-wire notification value and the
+    #: ``payload`` that :meth:`_split_packet` returns: ``ff09`` (2) + length (2) +
+    #: pattern (3) + cmd (2) + checksum (1). Added back so the fragmentation gate can
+    #: compare the notification length to the live ``ATT_MTU - 3`` cap.
+    _FRAME_OVERHEAD: int = 10
+
     def __init__(self, ble_device: BLEDevice) -> None:
         """Initialise device object. Does not connect automatically."""
 
         _LOGGER.debug(
             f"Initializing Solix device '{ble_device.name}' with"
-            f"address '{ble_device.address}' and details '{ble_device.details}'"
+            f"address '{ble_device.address}' and details '{ble_device.details}'",
         )
 
         self._ble_device: BLEDevice = ble_device
@@ -119,7 +125,6 @@ class SolixBLEDevice:
         self._connection_attempts = self._connection_attempts + 1
 
         try:
-
             # If we have an old client get rid of it
             if self._client is not None:
                 await self._dispose_of_client()
@@ -139,23 +144,24 @@ class SolixBLEDevice:
 
         except BleakError:
             _LOGGER.exception(
-                f"Error establishing initial connection to '{self.name}'!"
+                f"Error establishing initial connection to '{self.name}'!",
             )
 
         # If we are still not connected then we have failed
         if not self.connected:
             _LOGGER.error(
-                f"Failed to establish initial connection to '{self.name}' on attempt {self._connection_attempts}!"
+                f"Failed to establish initial connection to '{self.name}' on attempt {self._connection_attempts}!",
             )
             return False
 
         _LOGGER.debug(
-            f"Established initial connection to '{self.name}' on attempt {self._connection_attempts}!"
+            f"Established initial connection to '{self.name}' on attempt {self._connection_attempts}!",
         )
         try:
             _LOGGER.debug(f"Subscribing to notifications from device '{self.name}'!")
             await self._client.start_notify(
-                UUID_TELEMETRY, partial(self._process_notification, self._client)
+                UUID_TELEMETRY,
+                partial(self._process_notification, self._client),
             )
         except BleakError:
             _LOGGER.exception(f"Error subscribing/negotiating with '{self.name}'!")
@@ -164,10 +170,8 @@ class SolixBLEDevice:
         # Negotiate
         try:
             async with asyncio.timeout(NEGOTIATION_TIMEOUT):
-
                 # While negotiations have not completed
                 while not self.negotiated:
-
                     # If we have not received any packet from the device in
                     # any stage then restart negotiations from the start
                     if (
@@ -175,16 +179,15 @@ class SolixBLEDevice:
                         or (time.time() - self._last_packet_timestamp)
                         > NEGOTIATION_RESPONSE_TIMEOUT
                     ):
-
                         _LOGGER.debug(
-                            f"Sending negotiation initiation request to '{self.name}'..."
+                            f"Sending negotiation initiation request to '{self.name}'...",
                         )
                         await self._initiate_negotiations()
 
                     # Wait at this long to see if we get any response to
                     # our initial request in stage 0. This weird layout
                     # allows us to exit immediately when negotiation occurs
-                    for _ in range(0, NEGOTIATION_RESPONSE_TIMEOUT):
+                    for _ in range(NEGOTIATION_RESPONSE_TIMEOUT):
                         await asyncio.sleep(1)
                         if self.negotiated:
                             break
@@ -233,7 +236,6 @@ class SolixBLEDevice:
         for example, send a subscribe command to start a telemetry stream (see
         :class:`~SolixBLE.devices.c1000g2.C1000G2`).
         """
-        pass
 
     async def _keep_alive(self) -> int | None:
         """Execute designated keep-alive command periodically after good negotiation.
@@ -335,7 +337,11 @@ class SolixBLEDevice:
         return self._last_data_timestamp
 
     def _parse_int(
-        self, key: str, begin: int = None, end: int = None, signed: bool = False
+        self,
+        key: str,
+        begin: int = None,
+        end: int = None,
+        signed: bool = False,
     ) -> int:
         """Parse an integer at the specified key in the telemetry data.
 
@@ -379,23 +385,24 @@ class SolixBLEDevice:
 
         # Validate encoded length is correct
         packet_length = int.from_bytes(
-            bytes([packet_copy.pop(0), packet_copy.pop(0)]), byteorder="little"
+            bytes([packet_copy.pop(0), packet_copy.pop(0)]),
+            byteorder="little",
         )
         if packet_length != len(packet):
             raise ValueError(
-                f"Packet length is encoded as {packet_length} but its length was {len(packet)}!"
+                f"Packet length is encoded as {packet_length} but its length was {len(packet)}!",
             )
 
         # Validate checksum is correct
         packet_checksum = packet_copy.pop(-1).to_bytes()
         if packet_checksum != self._checksum(packet[:-1]):
             raise ValueError(
-                f"Packet checksum is encoded as {packet_checksum.hex()} but it is actually {self._checksum(packet[:-1]).hex()}!"
+                f"Packet checksum is encoded as {packet_checksum.hex()} but it is actually {self._checksum(packet[:-1]).hex()}!",
             )
 
         # Extract pattern
         packet_pattern = bytes(
-            [packet_copy.pop(0), packet_copy.pop(0), packet_copy.pop(0)]
+            [packet_copy.pop(0), packet_copy.pop(0), packet_copy.pop(0)],
         )
 
         # Extract command
@@ -469,17 +476,19 @@ class SolixBLEDevice:
 
                 # Sometimes there is just a param_id with no length or values
                 if len(remaining_data) == 0:
-                    parsed_data[param_id] = bytes()
+                    parsed_data[param_id] = b""
                     break
 
                 # Extract encoded length of parameter
                 param_len = int.from_bytes(
-                    _verbose_pop(remaining_data, 1, f"param_len (id={param_id})")
+                    _verbose_pop(remaining_data, 1, f"param_len (id={param_id})"),
                 )
 
                 # Extract data/body from parameter
                 param_data = _verbose_pop(
-                    remaining_data, param_len, f"param_data (id={param_id})"
+                    remaining_data,
+                    param_len,
+                    f"param_data (id={param_id})",
                 )
                 parsed_data[param_id] = param_data
 
@@ -487,13 +496,15 @@ class SolixBLEDevice:
                 _LOGGER.exception(
                     f"Unexpected end of packet! Data may be missing or invalid!"
                     f" Extracted so far: '{self._parameters_to_str(parsed_data)}'."
-                    f" Payload: '{payload.hex()}'"
+                    f" Payload: '{payload.hex()}'",
                 )
 
         return parsed_data
 
     def _parameters_to_str(
-        self, parameters: dict[str, bytes], types: bool = False
+        self,
+        parameters: dict[str, bytes],
+        types: bool = False,
     ) -> str:
         if types:
             with_types = {
@@ -506,8 +517,7 @@ class SolixBLEDevice:
                 for k, v in parameters.items()
             }
             return json.dumps(with_types, indent=4, sort_keys=True)
-        else:
-            return str({k: v.hex() for k, v in parameters.items()})
+        return str({k: v.hex() for k, v in parameters.items()})
 
     def _log_diff(self, old: dict[str, bytes], new: dict[str, bytes]) -> None:
         """Log any differences between parameters."""
@@ -522,13 +532,15 @@ class SolixBLEDevice:
             if new[k] != old[k]
         }
         _LOGGER.debug(
-            f"Parameter changes: \n{json.dumps(differences, indent=4, sort_keys=True)}"
+            f"Parameter changes: \n{json.dumps(differences, indent=4, sort_keys=True)}",
         )
 
     def _decrypt_payload(self, payload: bytes) -> bytes:
         """Decrypt telemetry packet using negotiated shared secret and IV."""
         cipher = AES.new(
-            self._shared_secret[:16], AES.MODE_CBC, iv=self._shared_secret[16:]
+            self._shared_secret[:16],
+            AES.MODE_CBC,
+            iv=self._shared_secret[16:],
         )
         decrypted = cipher.decrypt(payload)
         unpadder = PKCS7(128).unpadder()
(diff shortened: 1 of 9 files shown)


tests/test_reassembly.py
```diff
@@ -0,0 +1,70 @@
...
+CMD = b"\xc4\x90"
+
+
+def _device() -> SolixBLEDevice:
+    dev = SolixBLEDevice(MOCK_BLE_DEVICE)
+    dev._client = mock.Mock(mtu_size=256)
+    return dev
+
+
+def test_short_single_without_frag_byte_kept_whole() -> None:
```
flip-dots: Use test [parametrization](https://docs.pytest.org/en/stable/example/parametrize.html) here

SolixBLE/device.py
```diff
@@ -57,12 +57,18 @@ class SolixBLEDevice:
     #: (e.g the C1000 Gen 2 uses ``c421``/``c900`` instead of ``c402``/``c405``).
     _TELEMETRY_COMMANDS: tuple[str, ...] = ("c402", "4300", "c405")
 
+    #: Fixed ff09-frame overhead between the on-wire notification value and the
```
flip-dots: This comment is not very useful for someone who does not know thew context of this PR. It should ideally make sense on its own.

SolixBLE/device.py
```diff
@@ -57,12 +57,18 @@ class SolixBLEDevice:
...
+    #: compare the notification length to the live ``ATT_MTU - 3`` cap.
+    _FRAME_OVERHEAD: int = 10
+
     def __init__(self, ble_device: BLEDevice) -> None:
         """Initialise device object. Does not connect automatically."""
 
         _LOGGER.debug(
             f"Initializing Solix device '{ble_device.name}' with"
-            f"address '{ble_device.address}' and details '{ble_device.details}'"
+            f"address '{ble_device.address}' and details '{ble_device.details}'",
```
flip-dots, line 71: Ideally these unrelated formatting changes would not be present in the PR since it makes reviewing more difficult but I am fine with them being left in since the tooling complains if you don't and I have not properly formatted the codebase yet.

SolixBLE/device.py
```diff
@@ -514,59 +526,110 @@ def _encrypt_payload(self, payload: bytes) -> bytes:
...
-    ) -> None:
-        """Process a telemetry packet from the device.
-
-        This performs the default processing of telemetry packets in which
-        telemetry payloads are spread across multiple packets. This is
-        overridden for devices which do not use multi-packet payloads for
-        telemetry.
+    def _reassemble(self, cmd: bytes, payload: bytes) -> bytes | None:
+        """Reassemble a possibly-fragmented session frame into one payload.
+
```
flip-dots: This does a good job of explaining how the fragmentation works but not what it is.

I.e maybe start the main comment body with something like: 

Some Anker devices split the payload across multiple packets, this function is used to combine fragmented payloads or pass through non-fragmented payloads before they can be further processed and/or decrypted.

SolixBLE/device.py
```diff
@@ -514,59 +526,110 @@ def _encrypt_payload(self, payload: bytes) -> bytes:
...
+        run) is a fragment. A short frame is a standalone single, and it carries a
+        ``<index><total>`` byte *only* if that byte is a valid single marker
+        (``0x11``); some families (e.g. the A91B2 station) put no frag byte on
+        singles, so their first byte is already ciphertext and must be kept.
+
+        :returns: the complete (still-encrypted) payload ready to decrypt, or
+            ``None`` while fragments are still outstanding.
         """
+        if not payload:
+            return payload
```
flip-dots, line 587: Is this reachable?

SolixBLE/device.py
```diff
@@ -514,59 +526,110 @@ def _encrypt_payload(self, payload: bytes) -> bytes:
...
-        fragment_total = payload[0] & 0x0F
+        cmd_key = bytes(cmd)
+        index = (payload[0] >> 4) & 0x0F
+        total = payload[0] & 0x0F
 
-        # Multi-part message
-        if fragment_total > 1:
-            fragment_data = payload[1:]
-            cmd_key = bytes(cmd)
+        # Continuation (or short tail) of a run already in progress for this cmd.
```
flip-dots: Again I think this comment does not make sense outside of this PR, maybe it should be something like: if we already have fragments of this payload.

SolixBLE/device.py
```diff
@@ -514,59 +526,110 @@ def _encrypt_payload(self, payload: bytes) -> bytes:
...
+        if cmd_key in self._fragment_buffers:
             _LOGGER.debug(
-                f"Fragment {fragment_index}/{fragment_total} for cmd {cmd.hex()}, {len(fragment_data)} bytes"
+                f"Fragment {index}/{self._fragment_totals[cmd_key]} for cmd "
+                f"{cmd.hex()}, {len(payload) - 1} bytes",
             )
+            self._fragment_buffers[cmd_key][index] = payload[1:]
+            return self._join_fragments(cmd_key)
+
+        # A run only starts on a full-length first fragment. The length guard stops a
```
flip-dots: Ideally leave out specific terminology like a "run", something like: "subsequent fragments of a fragmented payload will always be preceded by a fragment which is the maximum size and this is used for x and the maximum size is determined by y" would get the same message across without needing to know additional context.

SolixBLE/device.py
```diff
@@ -514,59 +526,110 @@ def _encrypt_payload(self, payload: bytes) -> bytes:
...
+            and notification_length >= self._client.mtu_size - 3
+        ):
+            _LOGGER.debug(
+                f"Fragment 1/{total} for cmd {cmd.hex()}, {len(payload) - 1} bytes",
+            )
+            self._fragment_buffers[cmd_key] = {1: payload[1:]}
+            self._fragment_totals[cmd_key] = total
+            return self._join_fragments(cmd_key)
+
+        # Standalone single notification. Strip the frag byte only when it is a valid
```
flip-dots: Try to stick to pre-existing terminology like fragmented, non-fragmented, etc.

SolixBLE/device.py
```diff
@@ -514,59 +526,110 @@ def _encrypt_payload(self, payload: bytes) -> bytes:
...
+        # single marker; otherwise the whole payload is data.
+        if payload[0] == 0x11:
+            return payload[1:]
+        return payload
+
+    def _join_fragments(self, cmd_key: bytes) -> bytes | None:
+        """Join a completed fragment run, or ``None`` if more fragments are due."""
```
flip-dots: Try to avoid the use of run, something like: Return the merged payload if all fragments are present in the buffer and reset or return None if fragments are still missing.

SolixBLE/device.py
```diff
@@ -514,59 +526,110 @@ def _encrypt_payload(self, payload: bytes) -> bytes:
...
+        del self._fragment_buffers[cmd_key]
+        del self._fragment_totals[cmd_key]
+        _LOGGER.debug(f"Reassembled payload: {len(payload)} bytes")
+        return payload
 
-            # Wait until all fragments have arrived
-            if len(self._fragment_buffers[cmd_key]) < fragment_total:
-                _LOGGER.debug("Waiting for remaining fragments...")
-                return
+    async def _process_telemetry_packet(
```
flip-dots, line 668: I think it makes more sense to move the [_process_telemetry_packet()](https://github.com/flip-dots/SolixBLE/blob/f595493469469a7d3cb398f8e7d87295ade33a82/SolixBLE/prime_device.py#L490) override from prime_device.py to device.py. Catching exceptions here isn’t super useful.

SolixBLE/device.py
```diff
@@ -631,75 +695,75 @@ async def _process_notification(
...
-
                 # Non-encrypted telemetry messages
                 if cmd.hex() == "0300":
                     _LOGGER.debug("Received non-encrypted telemetry message!")
                     parameters = self._parse_payload(payload)
                     return await self._process_telemetry(parameters)
 
+                # Reassemble multi-fragment frames before the cipher, so telemetry
+                # and unknown session frames share one reassembler (SolixBLE #42).
+                payload = self._reassemble(cmd, payload)
```
flip-dots: Would it not make more sense to put the re-assembler before the payload is used anywhere? That way [_listen_for_packet()](https://github.com/flip-dots/SolixBLE/blob/f595493469469a7d3cb398f8e7d87295ade33a82/SolixBLE/device.py#L900) can be used to listen for fragmented payloads?

flip-dots (commented):
I have not tested this yet but it looks pretty good, I have just a few questions and improvements I would like to see to the tests and comments.

tests/helpers.py
```diff
@@ -105,19 +105,22 @@ def custom_init(*args, **kwargs):
...
-                name=f"bleak_client_{len(self._mock_bleak_clients)}"
+                name=f"bleak_client_{len(self._mock_bleak_clients)}",
             )
 
             # Set functions/properties
             mock_bleak_client.write_gatt_char.side_effect = self.write_gatt_char
             mock_bleak_client.start_notify.side_effect = self.start_notify
+            # Emulate an Anker 256-byte ATT MTU so fragment reassembly (which gates on
+            # the live ``mtu_size - 3`` notification cap) behaves as it does on device.
+            mock_bleak_client.mtu_size = 256
```
flip-dots, line 116: I think we are going to need tests for different MTU sizes, someone [reports](https://github.com/flip-dots/SolixBLE/issues/55#issuecomment-5103990681) the MTU size for their F2000 is 247 rather than 256.

kb1ibt:
Apologies for the delay — I was at Pennsic for the past 2 weeks with minimal connectivity, so I wasn't able to provide any responses to your review.

The branch is **rebased onto current `main`** (it had fallen 22 commits behind) and green — 113 tests, Read the Docs passing. Taking your points in order:

- **Test parametrization** — agreed, converting `test_reassembly.py`.
- **Comments that only make sense inside this PR** (`device.py` 60 and 564) and the **"run" terminology** (573, 589, 596) — all fair. Rewriting them to stand on their own and sticking to fragmented / non-fragmented throughout. Your suggested opening for the `_reassemble` docstring says what it *is* before how it works, which mine didn't; I'll use that shape.
- **Unrelated formatting churn** — acknowledged, and sorry for the extra review load. It's the linter reformatting lines adjacent to the ones I touched. I'll keep it out of the next push where I can.
- **Moving the `_process_telemetry_packet` override out of `prime_device.py`** — agreed on both counts; `device.py` is the right home and the exception handling there isn't earning its keep.
- **Reassembler placement** — you're right, and I notice you've already done exactly this in #61: reassembling before the futures dispatch so `_listen_for_packet` receives whole payloads rather than a first fragment. Mine currently sits after the futures check, which is wrong for the same reason. Moving it.
- **MTU sizes** — replying inline, because the F2000 report you linked points at something better than what either of us is doing.

One thing worth flagging given the timing: #57 landed while I was away, so the keep-alive machinery is yours now and the `_KEEPALIVE_CMD` approach in #51 is superseded. I'll rework that PR onto your `_keep_alive()` contract rather than competing with it.

tests/helpers.py
kb1ibt: Agreed, and that F2000 report is more useful than a second test case — it exposes that neither the current approach here nor a fixed constant is right.

`client.mtu_size` is unreliable on BlueZ. bleak's backend returns a **hardcoded 23** unless `_acquire_mtu()` has been called, with `UserWarning: Using default MTU value`. We hit this live on a Pi Zero 2W. Reassembly still works, because the comparison is `>=` and a real 253-byte fragment clears a threshold of 20 — but the guard silently stops doing its actual job, which is keeping a short frame whose first ciphertext byte happens to look like `0x1x` from opening a run that never completes. So on Linux, the platform HA runs on, it's effectively inert.

The device solves this for us: it **declares its own MTU** in the stage-2 response (`0803`/`4803`), field `a2` — `253` (`fd00`) on our A1783 / A91B2 / A2345, and `297` (`2901`) on a 160W, which lines up with the 247-MTU F2000 being a different number again. That value agrees with `ATT_MTU - 3` in every capture we have, and reading it needs no `_acquire_mtu()` call on any platform.

So I'll change this to prefer the declared `a2` and fall back to `mtu_size - 3`, and parametrize the tests over 247 / 253 / 297 rather than just emulating 256.

Worth noting the same field would fix the `_MAX_PACKET_SIZE = 253` constant in #61 — at 247 the cap is 244, so a `== 253` test never fires and fragments get treated as complete payloads.

SolixBLE/device.py
kb1ibt: Structurally yes, in practice no. A frame is `ff09` + length(2) + pattern(3) + cmd(2) + payload + checksum(1), so a 10-byte frame parses to an empty payload and `_split_packet` will hand one back. But I've never seen one on any of the three devices here, and none of the four captures contains one — even the "bare status" responses (`0805`, `0822`) carry a byte.

So it's purely a guard for `payload[0]` on the next line, protecting against a malformed or truncated notification rather than anything a device actually sends. Happy to drop it and let a bad frame raise if you'd rather the parser stay strict — it's the sort of defensive line that hides a real bug later.

tests/test_reassembly.py
kb1ibt: Folding the five cases (single with no frag byte, `0x11` single, two-fragment, exact-multiple-with-no-short-tail, cold index != 1) into one parametrized test, with the MTU as a parameter too so they run at 247 / 253 / 297 rather than only 256.

SolixBLE/device.py
kb1ibt: Replacing it with: the `ff09` frame is header(2) + length(2) + pattern(3) + cmd(2) + payload + checksum(1), so this is the 10 bytes of framing between a notification's on-wire length and its payload length — what you add to a payload to compare it against the notification cap.

SolixBLE/device.py
kb1ibt: These are ruff rather than a deliberate reformat — `[tool.ruff.lint] select = ["ALL"]` in `pyproject.toml` enables COM812 (missing trailing comma), and `.github/workflows/ruff.yml` gates the PR on it. Any call whose lines the change touches has to gain one for CI to pass.

SolixBLE/device.py
kb1ibt: Taking your opening more or less verbatim: some Anker devices split a payload across multiple packets, and this combines fragmented payloads or passes non-fragmented ones through before further processing or decryption.

Keeping one detail from the original: it runs before the cipher, which is what lets telemetry and unknown session frames share one reassembler whether the device is GCM or CBC.

SolixBLE/device.py
kb1ibt: Reworded to your suggestion: if we already have fragments of this payload.

SolixBLE/device.py
kb1ibt: Reworded to: subsequent fragments of a fragmented payload are always preceded by a fragment that fills the notification cap, so a full-length notification means more is coming and a shorter one closes it. The cap is the negotiated ATT MTU minus 3, which per the other thread should come from the MTU the device declares in its stage-2 response rather than from `client.mtu_size`.

SolixBLE/device.py
kb1ibt: Switched to fragmented / non-fragmented throughout the function, including this one.

SolixBLE/device.py
kb1ibt: Using your wording: return the merged payload if all fragments are present in the buffer and reset, or return None if fragments are still missing.

SolixBLE/device.py
kb1ibt: Moving it to `device.py`. The override only exists because Prime devices pack telemetry into a single frame, which the shared length gate now handles for every family, so there is nothing device-specific left in it.

The exception handling goes with it — it was catching a partial fragment that could not decrypt, and once the gate decides fragmentation by length rather than by trusting `payload[0]`, that case stops arising.

SolixBLE/device.py
kb1ibt: You are right. It currently sits inside the session-message branch, after the futures check, so a `_listen_for_packet` caller gets handed the first fragment rather than the whole payload — which is exactly the bug in `get_status_update()` on the C1000/C300/C800, where the code listens twice and stitches the two packets together by hand.

Moving it above the futures dispatch. I see #61 already puts it there and drops those manual two-packet stitches as a result.

SolixBLE/device.py
kb1ibt: Pushed in a128052, but I was wrong about half of this and want to correct it rather than leave it standing.

The override is gone — it only differed from the base by the guard around decryption, so there is now one implementation in `device.py`.

The exception handling could not go, though. I removed it, and four of your existing tests failed: `solix_packet_1_missing`, `solix_both_packets_reversed`, `solix_both_packets_later_out_of_order` all feed deliberately broken fragment sequences and assert `device._data` stays `None`, and without the guard the CBC decrypt raises `Data must be padded to 16 byte boundary` straight into the notification callback.

I then tried to make it unnecessary by having `_reassemble` drop a fragment that arrives with nothing to join it to, which is the more correct fix. That broke four *different* tests — `prime_telemetry_packet`, `prime_power_bank_telemetry_packet` and both MagGo cases. Those devices send non-fragmented payloads with no fragment header, so a ciphertext first byte that happens to read as a valid `<index><total>` is indistinguishable from a genuine stray fragment. Dropping one throws away real telemetry.

So the two cases can only be told apart by whether the payload decrypts, which means something has to absorb the failure. I have kept it where it was and rewritten the comment to say that, instead of implying it is defensive padding.

Worth noting you have solved this properly in #61 by wrapping the whole of `_process_notification` in a try/except — that is the right shape, and with it in place this guard could come out. It felt like scope creep to add the same thing here, but say the word and I will.

kb1ibt:
All of the work added in this PR was superseded by #61

Issue #17, opened 2026-03-14
Anker Prime 250W "Unexpected end of packet", sensors unavailable


pkolbus:
Hi, recently purchased an Anker Prime 250W charger and attempting to integrate into Home Assistant. Ran across this library and https://github.com/flip-dots/HaSolixBLE which seemed the most promising. When I add to Home Assistant (using [#18](https://github.com/flip-dots/HaSolixBLE/pull/18)) i can see the device and sensors but all ports are showing unavailable. (For what it's worth, the discovery/connection is via one of several ESPHome Bluetooth proxies).

(For what it's worth, https://github.com/thomluther/ha-anker-solix also does not work).

I see you've written some great documentation for adding/debugging a device; intend to to dive in to help debug/fix this. If you have any suggestions - things to look at or try -- these would be welcomed.

```
2026-03-14 19:06:20.374 ERROR (MainThread) [SolixBLE.device] Unexpected end of packet! Data may be missing or invalid! Payload: '00a10101044859710a627e653606c6d0f100b4e2'
Traceback (most recent call last):
  File "/usr/local/lib/python3.14/site-packages/SolixBLE/device.py", line 389, in _parse_payload
    param_data = bytes([remaining_data.pop(0) for _ in range(0, param_len)])
                        ~~~~~~~~~~~~~~~~~~^^^
IndexError: pop from empty bytearray
2026-03-14 19:06:20.475 ERROR (MainThread) [SolixBLE.device] Unexpected end of packet! Data may be missing or invalid! Payload: '00a10102a202fd00a30144a40101a501028c2e88c598023c7d27d6c9edddf0af9e'
Traceback (most recent call last):
  File "/usr/local/lib/python3.14/site-packages/SolixBLE/device.py", line 389, in _parse_payload
    param_data = bytes([remaining_data.pop(0) for _ in range(0, param_len)])
                        ~~~~~~~~~~~~~~~~~~^^^
IndexError: pop from empty bytearray
2026-03-14 19:06:20.578 ERROR (MainThread) [SolixBLE.device] Unexpected end of packet! Data may be missing or invalid! Payload: '00a10103a2054553503332a307302e302e302e33a41041514c514a4231463532323030303832a5067ce913833daab78ca884812ed76cba4e013df23dac65'
Traceback (most recent call last):
  File "/usr/local/lib/python3.14/site-packages/SolixBLE/device.py", line 389, in _parse_payload
[...]

PR #49, opened 2026-07-21
feat(ble): capability-driven encrypted negotiation + client-token auth for hardened firmware (#22)

kb1ibt:
Split out of #45 per review — the shared connection layer.

Replaces the fixed-replay negotiation frames with **live/dynamic** frames — each carries the current timestamp (and the stage-5 confer the local timezone), which newer firmware requires and rejects a stale one — and adds the account `owner_user_id` binding that hardened Prime devices need before they arm telemetry (without it they ack `09` and withhold updates).

This is what lets the **C1000 G2 / C2000 G2** and hardened Prime chargers complete negotiation and stream instead of being dropped mid-handshake — the disconnect reported in #22.

**Stacked on #48** (reassembly). The c490 summary decode, Prime device support, and docs follow as separate PRs on top of this one. 77 tests pass.


docs/source/owner_user_id.rst
```diff
@@ -0,0 +1,131 @@
...
+Method 1: anker-solix-api (recommended)
+---------------------------------------
+
+`anker-solix-api <https://github.com/thomluther/anker-solix-api>`_ is a Python client for
+the Anker cloud. Logging in once returns your ``user_id``. This needs only your Anker
+account credentials -- no phone, no rooted device.
+
+Requirements:
+
+- `anker-solix-api <https://pypi.org/project/anker-solix-api/>`_ (``pip install anker-solix-api``)
```
pkolbus: anker-solix-api is not published on PyPI, and installing from the git repo (git+https://github.com/thomluther/anker-solix-api) with pip or uv doesn't seem to install all dependencies. It does look like there is some poetry support in the pyproject.toml though.

docs/source/owner_user_id.rst
```diff
@@ -0,0 +1,131 @@
...
+- Your Anker account email and password
+
+.. code-block:: python
+
+    import asyncio
+
+    from aiohttp import ClientSession
+    from api.api import AnkerSolixApi  # anker-solix-api
```
pkolbus:
```suggestion
    from anker_solix_api.api import AnkerSolixApi
```

docs/source/owner_user_id.rst
```diff
@@ -0,0 +1,131 @@
...
+
+
+    async def main() -> None:
+        async with ClientSession() as session:
+            api = AnkerSolixApi("EMAIL", "PASSWORD", "US", session)
```
pkolbus, line 74: Users outside the US may benefit from a bit of discussion about how the countryId should be set... 

https://github.com/thomluther/anker-solix-api/blob/9a57579d468281395098ff55c8c61b6541cf9596/src/anker_solix_api/apitypes.py#L23 or https://github.com/search?q=repo%3Athomluther/anker-solix-api%20API_COUNTRIES&type=code

pkolbus (commented):
@kb1ibt, nice work!

@flip-dots this (PR stack) gets my 250W charging station working -- addressing the problem we tried to work through in #17. I'm guessing that it's this PR in particular, although the others probably help.

Some comments on the owner-id collection here.

I did take a pass at improving the script snippet in owner_user_id.rst, as https://gist.github.com/pkolbus/7cdb87b04d6673ae93218ac4cf90b818. [uv](https://github.com/astral-sh/uv) provides a way to run a script in an isolated env with declared dependencies, and the Anker credentials and country ID are collected as terminal input. It's probably fine to have this collection remain separate and owner ID be a configuration item on the HA side, at least initially. anker_solix_api doesn't have any published packages on PyPI and keeping the script separate reduces the dependency burden and keeps the integration clearly cloud-free.

kb1ibt:
Thanks — all three were right, and one was a real copy-paste breaker. Fixed in `8c9d82f`:

- **PyPI:** confirmed, `pypi.org/pypi/anker-solix-api` 404s, so both the `pip install` line and the link were wrong. Now documents clone + `poetry install`, plus your `uv` inline-metadata approach (with the credential prompts) as the least invasive way to run it once — nothing lands in system Python.
- **Import:** `from api.api import …` doesn't resolve at all; the package is `anker_solix_api`. Applied your suggestion.
- **countryId:** added a section noting it selects the API server region, so a wrong code fails the login outright rather than returning something different, pointing at `API_COUNTRIES` for valid values.

Also noticed while building the docs that the page wasn't in any toctree, so Sphinx warned and it was unreachable by navigation — added it in the same commit.

Agreed it should stay separate, and for a slightly broader reason than HA: SolixBLE gets consumed outside Home Assistant too. Mine is a standalone collector that ships both cloud-MQTT and local-BLE telemetry into InfluxDB line protocol — it takes `owner_user_id` from its own config file, fetching the account `user_id` from the Anker login once and writing it back to config, then running BLE-only in remote locations like Pennsic or Burning Man with no further cloud round-trip. That's really the point of keeping collection out of the library: you obtain the ID once where there's internet, and everything after that is offline. Leaving it as a caller-supplied config value works for any consumer, HA or not.

On #17 — glad the stack fixed it. For the record, the `Unexpected end of packet` had a specific cause: those payloads parse cleanly as TLVs and then hit exactly 16 trailing bytes, which is the AES-GCM authentication tag being decrypted with the body instead of stripped. The TLV walker reads the tag's first byte as a tag/length pair and pops an empty bytearray. Not corruption — a MAC-tag handling bug, which is why it looked like truncation.

kb1ibt:
Rebased onto `main` (post-#61 packet rewrite). Changes from the prior version: negotiation is now capability-driven — the advert's capability byte (0x04 ECDH bit) picks the encrypted `4xxx`/GCM vs cleartext `0xxx`/CBC path before connect; the account `owner_user_id` binding is replaced with a self-generated, persisted **client token** enrolled once by a physical button press (no cloud/account dependency); reworked onto the `construct`-based `Packet`/`Parameters` layer.

kb1ibt:
This is 1/2 of the work needed for #68

kb1ibt:
@flip-dots can you re-review this since it is part of the blocker for #66 / #68 and likely quite a few others.

kb1ibt:
Rebuilt from `main` on your review and split into two focused PRs: #70 (negotiation path, chosen from the advertised capability byte) and #71 (client-token pairing with button-press confirmation). Closing this in favour of those.

PR #50, opened 2026-07-21
feat(ble): decode the c490 protobuf device-summary (C1000 G2 / C2000 G2)

kb1ibt:
Split out of #45 per review — the c490 protobuf device-summary decode.

Adds `walk_protobuf` (`parsing.py`) + `_protobuf_body` to decode the protobuf device-summary frame (the C2000 G2's `c490`) into a `.path` field map exposed via `summary`, rather than mis-reading it as the 1-byte-tag TLV the base parser expects (which garbles the varint tags). Adds the **C2000 G2 (A1783)** device on top of the C1000 G2, and the documentation for it.

**Rebased onto current `main`** and decoupled from #49 — this PR no longer carries the negotiation work. It is now just **#48 (reassembly) + the c490 decode + its docs**, three commits. The reassembly commit is included because the c490 frame is multi-fragment and does not decode without it; everything else is this PR's own.

Per the review note on #52, the C2000 G2 documentation now lives here rather than in a separate docs PR: `docs/source/c2000g2.rst`, its `api.rst` toctree entry, and a C2000 G2 column in the power station support table with the rows this decode makes readable (max charge power, pack voltage, cumulative energy out, charge presence). Only the A1783 was verified against hardware; the A1763 column reflects the shared code path.

142 tests pass; Sphinx builds clean.


SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class C2000G2(C1000G2):
+    """
+    C2000(X) Gen 2 Power Station.
+
+    Use this class to connect, monitor and control a Gen 2 C2000(X) power
+    station. This model is also known as the A1783.
+
+    The C2000 G2 is the larger sibling of the C1000 G2 (A1763) and shares its
```
flip-dots, line 18: Technical details like this (anything which starts talking about implementation rather than what you need to know to use the module) which are automatically rendered in the docs should be hidden behind a collapsed note like [here](https://github.com/flip-dots/SolixBLE/blob/cb684b333467e7b1582509a67770223af549564a/SolixBLE/devices/maggo_3in1.py#L26-L27)

SolixBLE/device.py
```diff
@@ -98,11 +114,50 @@ def remove_callback(self, function: Callable[[], None]) -> None:
...
+    @staticmethod
+    def _ts() -> str:
+        """Current unix time as a 4-byte little-endian hex string.
+
+        Used as the live replay-protection timestamp in the negotiation and session
+        commands; newer firmware rejects a stale (frozen) value and drops the link.
+        """
+        return int(time.time()).to_bytes(4, "little").hex()
+
+    @staticmethod
```
flip-dots: This code sucks but I have had a quick look around and there does not seem to be a better way to do it. Maybe create a utils.py file and hide this code inside it, device.py is already large/complicated enough.

flip-dots: I shouldn’t be surprised that even in 2026 time zones are still awful to deal with.

SolixBLE/device.py
```diff
@@ -305,8 +356,24 @@ def last_update(self) -> datetime | None:
         """
         return self._last_data_timestamp
 
+    @property
+    def summary(self) -> dict[str, object]:
+        """Fields from the latest protobuf device-summary frame, if any.
+
+        Populated from a ``_PROTOBUF_TELEMETRY_COMMANDS`` frame (e.g. the C2000 G2's
```
flip-dots, line 352: This is probably fine for now but I might end up moving this stuff into a dedicated class which supporting devices can implement to get this functionality depending upon how widespread this bonus telemetry ends up being.

SolixBLE/device.py
```diff
@@ -708,56 +873,60 @@ async def _process_negotiation(self, cmd: bytes, payload: bytes) -> None:
...
-                    "Entered negotiation stage 1 due to response from device!"
+                    "Entered negotiation stage 1 due to response from device!",
                 )
                 parameters = self._parse_payload(payload)
                 _LOGGER.debug(f"Parameters: {self._parameters_to_str(parameters)}")
                 _LOGGER.debug("Sending stage 1 response message...")
                 return await self._client.write_gatt_char(
-                    UUID_COMMAND, bytes.fromhex(NEGOTIATION_COMMAND_1)
+                    UUID_COMMAND,
+                    self._negotiation_packet("0003", "a30120a40200f0"),
```
flip-dots: Use constants for these and other magic values like them.

SolixBLE/prime_device.py
```diff
@@ -49,23 +50,39 @@
     "a104f079b569a30400000000a518474d54304253542c4d332e352e302f312c4d31302e352e30"
 )
 
+#: Account owner_user_id bound into the 4027 registration (``a228``). Hardened
+#: (SolixBLE #22) firmware rejects a plain ``a224<uuid>`` registration (ack status
+#: ``09``) and only arms telemetry once the registration carries the account that
+#: owns the device. Account-specific -- MUST be sourced from config, never hard-coded
+#: upstream. ``None`` falls back to the legacy ``a224<uuid>`` payload.
```
flip-dots: Do we know what happens if the device is setup with Bluetooth only (i.e not setup using the cloud)? 

It would be good to have a way to use these newer devices without the need for the cloud at all.

That being said I think it still makes sense to support this for devices like the Solarbank where you might want both cloud and local control.

SolixBLE/prime_device.py
```diff
@@ -164,24 +193,28 @@ def _decrypt_payload(self, payload: bytes) -> bytes:
     # Negotiation #
     ###############
 
-    async def _initiate_negotiations(self) -> None:
-        """
-        Send the negotiation initiation command.
-        """
+    def _live_negotiation_packet(self, cmd: str, extra: str = "") -> bytes:
```
flip-dots: Why not reuse _negotiation_packet() from device.py?

SolixBLE/prime_device.py
```diff
@@ -201,111 +233,82 @@ async def _process_negotiation(self, cmd: bytes, payload: bytes) -> None:
...
-                    "Entered negotiation stage 3 due to response from device!"
+                    "Entered negotiation stage 3 due to response from device!",
                 )
                 decrypted_payload = self._decrypt_payload(payload)
                 _LOGGER.debug(f"Decrypted payload: {decrypted_payload.hex()}")
                 parameters = self._parse_payload(decrypted_payload)
+                # Device identity (chip, BLE firmware, serial, MAC) is sent only
+                # in this stage-3 message, never in telemetry, so persist it for
+                # the identity properties to read.
+                self._device_info = parameters
```
flip-dots: I think self._device_info and self._summary should be named self._data_device and self._data_summary to be consistent with the existing self._data and distinguish it from other cached values.

flip-dots (commented):
The more I read this PR the more I think this needs to be in its own dedicated class (maybe a new type e.g cloud_device.py) since its quite different from the way a lot of other devices work but I am not entirely decided about that. 

Is this cloud stuff in the negotiations 100% needed to make this device work on the latest firmware? How does the app handle local usage only? Is there a reset command we can use to use our own cloud ID or some sort of adoption process we can use?

I am not completely against allowing this library to work with a device already set up with the cloud (especially for the solar banks for which its more or less mandatory) but if it can be avoided (which I am hoping it can be) I would rather we make sure there is nothing that can be done there first before implementing this cloud protocol that is overly difficult for an end user to use.

SolixBLE/prime_device.py
kb1ibt: > Do we know what happens if the device is setup with Bluetooth only (i.e not setup using the cloud)?

Now we do — I unsealed two A2345s and tested it. **The `owner_user_id` path fails silently on an unowned device, but there is a cloud-free path, and it's the same opcode.**

**Never bound to any account:** `4027` carrying a 40-byte `a2` = account id gets **no reply at all**. Not `09`, not `04` — nothing. The session stays open, `4200`/`420a`/`420b` go out, `connect()` reports success, and the device simply never streams, dropping the link ~40 s later. Indistinguishable from a missing keepalive or a flaky link — and it is exactly what #22 looks like (`NEGOTIATED: True`, every field `-1`).

**The cloud-free path:** a **16-byte token in field 2** instead of the 40-byte account id.

| `4027` field 2 | response | then |
|---|---|---|
| 40 B account id, device **owned by that account** | `00` | streams (this stack — confirmed by @pkolbus on #17's hardware) |
| 40 B account id, device owned by **nobody** | *silence* | never streams |
| **16 B token, new to the device** | `09` + `a1 02 1e00` (30) | press the side button → device pushes an unsolicited `4827 00` → streams |
| **16 B token, already enrolled** | `00` | streams, no button |

The `09` is not a refusal — it means *awaiting physical confirmation*. The session stays up and the grant arrives asynchronously 5–8 s later when the button is pressed. **It persists**: reconnecting with the same token gets `00` immediately thereafter.

The token content is not checked against the device — I enrolled one unit using the *other* unit's serial, then `ABCDEFGHIJKLMNO`. But it is not ignored either: each **new** token needs its own button press, including on an already-paired unit. So it is a real pairing model — one press per client, and enrolling one client does not open the device to the next.

Reproduced identically on both units (`AQLQJB1G22202385`, `AQLQJB1G22202305`), so this is model behaviour, not a per-unit quirk.

**Inference:** field 2's *length* selects the mechanism — `0x28` (40) = account binding, `0x10` (16) = local client token. That fits your point about still wanting cloud support for devices like the Solarbank: both live on the same opcode, and a device can support either.

**One library gap this exposes.** The grant arrives on pattern **`030101`** — neither `030001` nor `03010f` — so the library logs

```
Unexpected packet type b'\x03\x01\x01' sent by device!
```

and discards it, which is the single most important frame in the exchange. Handling `030101`, and surfacing the `09`/`1e00` state so a UI can prompt "press the button on the device", is the minimum needed for cloud-free setup to work at all.

kb1ibt: I didn't realize the mistake, but `ABCDEFGHIJKLMNO` is only 15, not 16, so it even accepted a non-well-formed/truncated TLV.

SolixBLE/prime_device.py
kb1ibt: No good reason — I missed that `device.py` already had `_negotiation_packet()`. The two are identical apart from wrapping the payload in `_encrypt_payload`, so it wants a hook on the base rather than a near-duplicate.

Worth flagging where it lives, though: this line isn't #50's. It's from #49 (pr/negotiations), which #50 was stacked on when you reviewed it — I can't open a PR whose base is another of my branches in your repo, so the diff carried #49's changes along. Since the rebase, #50 doesn't touch `prime_device.py` at all.

Which makes the fix a question rather than a promise: given you're planning to merge #61 instead of #48 and build the dynamic timestamp/timezone work on top of it, that whole method may be replaced anyway. Happy to unify it in #49 if #49 still has a future, or to leave it alone if #61 subsumes it — your call, and I'd rather not spend your review time on code you're about to delete.

kb1ibt:
Rebased onto `main` (post-#61). c490 protobuf device-summary decode surfaced via `summary`/`summary_schema`, guarded by a schema-version check; adds the C2000 G2 (A1783) device; folds in the support-table rows from the closed #52 docs PR.
1.2.1.6 / 1.9.3.3 changed which map they are pulling, it is now _0009; but I haven't finished verifying every field in it vs _0005.

kb1ibt:
Rebuilt from `main` on your review as #72 — the C2000 G2 (A1783) with the c490 device-summary decode. Closing this in favour of it.

PR #51, opened 2026-07-21
feat(prime): A91B2 (240W) + A2345 (250W) live BLE telemetry + control

kb1ibt:
Split out of #45 per review — live BLE telemetry for the Anker Prime chargers.

Adds the Prime USB-charger base (`prime_usb_charger`) carrying the shared `4303`/`ca00` telemetry decode, and builds the **A2345** (250W charger) and **A91B2** (240W charging station) devices on it: per-port voltage/current/power/status, total output, per-port switches, and the station's two-frame-layout / AC-switch specifics.

**Stacked on #50** (c490) → #49 (negotiation) → #48 (reassembly). GitHub shows the cumulative diff until those merge; this PR's own contribution is the last commit (the Prime device support). 123 tests pass.


SolixBLE/devices/prime_charger_250w.py
```diff
@@ -4,275 +4,95 @@
...
+            + self.usb_c4_power
+            + self.usb_a1_power
+            + self.usb_a2_power,
+            2,
+        )
 
     @property
-    def usb_c2_current(self) -> float:
-        """USB C2 Port current (A).
+    def usb_c1_switch(self) -> bool:
```
flip-dots, line 46: Is it not possible to use PortStatus like the other devices?

SolixBLE/devices/prime_charger_250w.py
```diff
@@ -4,275 +4,95 @@
...
-            return DEFAULT_METADATA_FLOAT
+            return DEFAULT_METADATA_BOOL
 
-        return self._parse_int("a3", begin=4, end=6) / 1000.0
+        return bool(self._parse_int("aa", begin=1, end=2))
 
     @property
-    def usb_c2_power(self) -> float:
-        """USB C2 Port power (W).
+    def usb_c2_switch(self) -> bool:
```
flip-dots, line 57: Needs to use the same naming scheme as other devices. e.g [here](https://github.com/flip-dots/SolixBLE/blob/cb684b333467e7b1582509a67770223af549564a/SolixBLE/devices/prime_charger_160w.py#L118)

SolixBLE/devices/prime_usb_charger.py
```diff
@@ -0,0 +1,341 @@
...
+
+    @property
+    def usb_port_c1(self) -> PortStatus:
+        """USB C1 Port Status.
+
+        :returns: Status of the USB C1 port.
+        """
+        return PortStatus(self._parse_int("a4", begin=1, end=2))
+
+    @property
```
flip-dots, line 105: Device properties (e.g power, voltage, etc) should only be on the device class itself (e.g C300.py) not in a class inherited by others (e.g device.py / prime_device.py).

Hard coding port selections into a class which might be inherited by multiple devices prevents you from using it with future devices which may not have those ports. E.g prime_charger_160w is unable to use this class since it only has 3 USB C ports and no type A ports.

SolixBLE/devices/prime_charging_station_240w.py
```diff
@@ -0,0 +1,272 @@
...
+    UUID_COMMAND,
+)
+from ..device import SolixBLEDevice
+from ..prime_device import PRIVATE_KEY
+from .prime_usb_charger import PrimeUsbCharger
+
+_LOGGER = logging.getLogger(__name__)
+
+#: Cleartext-negotiation / confer packet pattern (``0xxx`` and ``4022``/``4023``).
+_NEGOTIATION_PATTERN = b"\x03\x00\x01"
```
flip-dots, line 31: Use the same format as [other constants](https://github.com/flip-dots/SolixBLE/blob/cb684b333467e7b1582509a67770223af549564a/SolixBLE/const.py#L54).

SolixBLE/devices/prime_charging_station_240w.py
```diff
@@ -0,0 +1,272 @@
...
+#: Encrypted session (data command) packet pattern.
+_SESSION_PATTERN = b"\x03\x00\x0f"
+
+
+class PrimeChargingStation240w(PrimeUsbCharger):
+    """Anker Prime Charging Station (240W / A91B2), an 8-in-1 charging station.
+
+    Despite sharing the Prime USB-charger telemetry layout, the station is **not**
```
flip-dots, line 39: Technical details not needed to simply use the library should be hidden in a collapsed note.

SolixBLE/devices/prime_charging_station_240w.py
```diff
@@ -0,0 +1,272 @@
...
+    a Prime/GCM device -- it is a base/**CBC**-lineage device (like the C1000 Gen 2):
+    it negotiates *in the clear* (``0xxx``) rather than with the encrypted ``4xxx``
```
flip-dots, line 41: Does using a base SolixBLEDevice class work?

SolixBLE/devices/prime_charging_station_240w.py
```diff
@@ -0,0 +1,272 @@
...
+_LOGGER = logging.getLogger(__name__)
+
+#: Cleartext-negotiation / confer packet pattern (``0xxx`` and ``4022``/``4023``).
+_NEGOTIATION_PATTERN = b"\x03\x00\x01"
+#: Encrypted session (data command) packet pattern.
+_SESSION_PATTERN = b"\x03\x00\x0f"
+
+
+class PrimeChargingStation240w(PrimeUsbCharger):
+    """Anker Prime Charging Station (240W / A91B2), an 8-in-1 charging station.
```
flip-dots, line 37: Damn all the cool Anker stuff seems to be for the US market only, it would be cool as hell if there was a UK/EU version of this.

SolixBLE/devices/prime_charging_station_240w.py
```diff
@@ -0,0 +1,272 @@
...
+        the base connect/reconnect machinery while replacing only the negotiation.
+        """
+        private_key = derive_private_key(int(PRIVATE_KEY, 16), SECP256R1())
+        public_key = private_key.public_key().public_bytes(
+            Encoding.X962,
+            PublicFormat.UncompressedPoint,
+        )[1:]
+
+        stages = (
+            ("0001", "a104" + self._ts(), "0801"),
```
flip-dots, line 123: This looks suspiciously similar to the negotiation for the regular Solix devices. Does treating it as a Solix rather than a Prime type device work?

SolixBLE/devices/prime_charging_station_240w.py
```diff
@@ -0,0 +1,272 @@
...
+
+        Runs on every (re)connection once the session is negotiated. Fire-and-forget:
+        ``4822``/``4823`` confer acks are harmless (see :meth:`_process_negotiation`),
+        and ``4a00``/``4303`` responses flow through the telemetry path.
+        """
+        serial = (self._device_info or {}).get("a4", b"")
+        timezone = self._local_posix_tz().encode()
+
+        # 4022 -- timezone; 4023 -- bind device serial (both AES-CBC, 030001).
+        await self._send_packet(
```
flip-dots, line 181: Given the similarity to the Solix style devices which usually don't need polling to provide any info, is this needed?

SolixBLE/devices/prime_charging_station_240w.py
```diff
@@ -0,0 +1,272 @@
...
+            raise
+        except Exception:
+            _LOGGER.debug(
+                "A91B2 '%s' realtime keep-alive stopped", self.name, exc_info=True
+            )
+
+    # ------------------------------------------------- 4a00 snapshot additions
+
+    @property
+    def ac_1_switch(self) -> bool:
```
flip-dots, line 240: Use the same property names as existing ones from other devices. There is already ac_output and since this is the first device with two AC outputs I think it makes sense to call them ac_output_1 and ac_output_2. Try to stick to convention for the others as well since the Home Assistant integration which uses this library relies on different devices using the same names for ports.

SolixBLE/devices/prime_charging_station_240w.py
```diff
@@ -0,0 +1,272 @@
...
+        """AC outlet 2 switch state (from the ``4a00`` snapshot).
+
+        :returns: True if on, else False (or default bool if no data).
+        """
+        if not self._data or "ab" not in self._data:
+            return DEFAULT_METADATA_BOOL
+        return bool(self._parse_int("ab", begin=1, end=2))
+
+    @property
+    def usb_total_power_out(self) -> float:
```
flip-dots, line 260: If you really want this property we can keep it but I would rather avoid adding more properties that don't already exist (e.g total_power_out is already established) since its more maintenance and testing and its easy enough for someone to aggregate the values themselves.

flip-dots (commented):
I am not certain but I suspect that these devices would work if you just changed the base class from PrimeDevice to SolixBLEDevice for them since the negotiations look very similar to that of earlier Solix devices and that protocol is implemented in the base SolixBLEDevice. 

Other than that try to stick to pre-existing convention, like properties not being inherited, using existing types, and using the same property names.

SolixBLE/devices/prime_charging_station_240w.py
kb1ibt: Your instinct here and on the negotiation below (*"this looks suspiciously similar to the negotiation for the regular Solix devices"*) is right, and the device says so out loud before you connect to it.

The advertisement's `0xffff` manufacturer record ends in a **capability byte**, and it declares which negotiation the unit accepts:

| device | capability | negotiation |
|---|---|---|
| **A91B2 station** | **`00`** | cleartext `0xxx` — the regular Solix path |
| A2345 charger | `04` | encrypted `4xxx` under the static pre-shared key |
| C2000 G2 (A1783) | `04` | encrypted `4xxx` |

So the A91B2 isn't "a Prime device with a Solix-looking handshake" — it is a **cleartext-only device**, which is exactly the base `SolixBLEDevice` behaviour, and that's why the negotiation looked familiar. The device even echoes the same fact on the wire as `authMethod` (field 5) in its stage-2 response: `00` = cleartext, `02` = encrypted.

That suggests the split isn't really Prime-vs-Solix at all — **it's which auth path the unit supports**, which is readable at scan time and confirmed again during negotiation. Selecting on that would put the A91B2 on the base class by rule rather than by exception, and keep the Prime class for the `04`/static-key units.

One caveat so this isn't a blanket yes: **negotiation** is the part that matches. The station still differs in telemetry — the two-frame layout and the AC-switch handling — so a straight base-class swap fixes the handshake but not the decode. If the class split were reorganised around auth path, the station would want the base negotiation plus its own telemetry parsing, rather than inheriting the Prime device wholesale.

Worth noting the same byte flags something in the other direction: **the C2000 G2 advertises `04`, and the library currently negotiates it in plaintext `0xxx`.** Both work — the app uses `4xxx` against that same unit — so nothing is broken today, but it does mean the library is choosing the weaker path on hardware that advertises support for the stronger one. I've put the full advert layout and samples on #61 since it touches the parsing work there.

SolixBLE/devices/prime_charging_station_240w.py
pkolbus: Not sure where this comment stands in terms of implementation, but total power (and energy) sensors are very useful for the HA energy dashboard among other things. I've set up group and integral helpers, and given the prominence that energy has, I'd hazard a guess that I'm not the only one. While not hard for an experienced user to create, it's more thing to do multiplied by every user that wants it, and may be a little tricky for users new to Home Assistant. I'd gently suggest taking overall value into account.

flip-dots: I think it makes more sense to implement aggregate sensors in the HA integration than here, that way it avoids any confusion between values reported by the device and ones generated from values reported by the device, since not all devices directly report things like total power out. 

That distinction is perhaps a bit pedantic but SolixBLE is targeted at developers, where that might be important, wheras HaSolixBLE is targeted at users, who probably don't care as long as the sensor exists.

kb1ibt:
Rebased onto `main` (post-#61). The A2345 charger and frame reassembly it originally carried are now upstream, so this is scoped to the **A91B2 charging station**: a base/CBC device on `SolixBLEDevice` (capability-driven crypto); per-port on/off + auto-off timers matching the 250w grammar (turn_ac_1_on, turn_usb_c1_on, set_timer_*); USB-A is telemetry-only.

SolixBLE/devices/prime_charging_station_240w.py
kb1ibt: You're right that the Solix-style devices stream without polling, and — as in the capability-byte thread above — the A91B2 is one of them for *negotiation*. Where it differs is telemetry, and this is the concrete case: its real-time feed is a **latched, self-expiring stream**, not a persistent push.

`420b` arms the `4303` stream, but the station only pushes for a bounded window. In the decompiled MCU firmware, the real-time trigger (`cmd20b`) loads an internal counter with 10 and decrements it per tick, so `4303` goes quiet ~8–10s after each arm. So it's not polling to fetch state (the connect-time `4a00` snapshot already does that); it's re-arming a stream that shuts itself off.

Without it, `_last_packet_timestamp` goes stale and any staleness-based consumer (e.g. the HA coordinator) treats the device as dropped and re-runs `_post_connect` — which re-sends the one-time confer (`4022`/`4023`) and re-binds the serial every cycle. `_keep_alive` just re-issues `4200` + `420b` on its interval to keep the feed live; it deliberately does **not** re-send the confer (that's why connect is split into `_post_connect` = confer-once vs `_request_stream` = the re-armable `4200`/`420b`, and the keep-alive only calls the latter).

It's the same base `_keep_alive` hook, and the same reason the C2000 G2 uses it: there the `4057` real-time latch can go dead for 20s+ between `c421` pushes, so a `4100` re-poll bridges the gap. Same pattern here (the `420b`/`4303` latch lapsing, just a shorter ~8-10s window) — so it stays device-specific and off the path of the Solix devices that genuinely don't need it.

kb1ibt:
Because `4303` and `4a00`/`ca00` use different layouts, having both in _TELEMETRY_COMMANDS caused the data to mismatch between them. So, making some heavy changes to your A2345 class to handle the two different maps.

docs/source/index.rst
```diff
@@ -29,50 +29,65 @@ No pairing is required in order to receive telemetry data or control the device.
...
+AC Power out            ✅      N/A        ✅      ✅       ✅       ✅       ✅          ✅    ✅
+AC on/off state         ✅      N/A        ✅      ✅       ✅       ✅       ❌          ✅    ✅
+AC Timer                ✅      N/A        ✅      ✅       ❌       ✅       ❌          ✅    ❌
+DC on/off control       ✅      ✅         ✅      ✅       ✅       ✅       ❌          ✅    ✅
+DC Power in             ✅      ✅         ✅      ✅       ✅       ✅       ✅          ✅    ✅
+DC Power out            ✅      ✅         ❌      ✅       ✅       ✅       ✅          ✅    ✅
+DC Power in status      ✅      ✅         ❌      ❌       ✅       ✅       ❌          ✅    ❌
+DC Power out status     ✅      ❌         ❌      ✅       ✅       ✅       ❌          ✅    ✅
+DC Timer                ✅      ✅         ❌      ❌       ❌       ✅       ❌          ✅    ❌
+Max charge power        ❔      ❔         ❔      ❔       ❔      ✅       ❔          ❔    ❔
```
flip-dots, line 62: Stick to the convention of green tick and red cross here, while I do think the table could be improved by adding more symbols and I would be fine accepting a sensible PR which implements it, it should not be part of this PR.

docs/source/index.rst
```diff
@@ -29,50 +29,65 @@ No pairing is required in order to receive telemetry data or control the device.
...
+Serial number           ✅      ✅         ✅      ✅       ✅       ✅       ✅          ✅    ✅
+Expansion temperature   N/A     N/A        N/A     ✅       N/A      ✅       ✅          ✅    ❌
+Expansion percentage    N/A     N/A        N/A     ✅       N/A      ✅       ✅          ✅    ✅
+Expansion health        N/A     N/A        N/A     ✅       N/A      ✅       ✅          ✅    ❌
+Expansion firmware      N/A     N/A        N/A     ✅       N/A      ✅       ✅          ✅    ✅
+Expansion num           N/A     N/A        N/A     ✅       N/A      ✅       ✅          ✅    ❌
+Polled status updates   ✅      ❌         ✅      ✅       ❌       ✅       ❌          ✅    ❌
+======================= ======= ========== ======= ======== ======== ======== =========== ===== =====
+
+The C2000 G2 (A1783) ``Pack voltage``, ``Cumulative energy out`` and ``Charge
```
flip-dots, line 86: Notes/warnings like this which are device specific should be in the device specific page, not the overall one.

docs/source/index.rst
```diff
@@ -131,22 +146,30 @@ Expansion battery serial number     ❌           ❌
...
+Individual port voltage ✅            ✅            ✅            ❌
+Individual port current ✅            ✅            ✅            ❌
+Individual port power   ✅            ✅            ✅            ✅
+AC outlet state         N/A           ✅            N/A           N/A
+Temperature             ❌            ❌            ❌            ❌
+Firmware version        ❌            ✅            ❌            ❌
+Serial number           ❌            ✅            ❌            ❌
+======================= ============= ============= ============= ===================
+
+The 240w (A91B2) is an 8-in-1 charging station: it decodes the same per-USB-port
```
flip-dots, line 167: Notes/warnings like this which are device specific should be in the device specific page, not the overall one.

docs/source/encrypted_negotiation.rst
```diff
@@ -0,0 +1,64 @@
+=================================
```
flip-dots, line 1: I think there should be a page or pages in the docs about the different protocols but I don’t think this is the way to organise it. 

There are multiple variants of the protocol, some use different forms of encryption, some use account binding, some are plain text, some need a button to be pressed, and some are a mix of encryption and plain text, and the earliest uses an entirely different packet structure. 

I don't expect you to write all of that but maybe we structure it as a protocols page which has a section for each type, (i.e one for plain text, one for CBC one for GCM, one for solar bank) and at the top of the page we can have the details about which paths are used?

SolixBLE/device.py
```diff
@@ -780,6 +1077,225 @@ async def _process_negotiation(self, cmd: bytes, payload: bytes) -> None:
...
+                )
+
+            # Stage 7: authorization result.
+            case "4827":
+                if plaintext[:1] == b"\x00":
+                    _LOGGER.debug("Client registration accepted; link authorized!")
+                    self._authorized = True
+                elif plaintext[:1] == b"\x09":
+                    _LOGGER.info(
+                        "Device is awaiting a physical button press to authorize "
```
flip-dots, line 1263: I think the library needs to respond in a more concrete way than just logging a message to tell you that you need to press the button, though what form that takes I am not sure, perhaps connect should raise a custom exception to tell the user they need to press the button and then they call connect again once they have? or maybe allow for a callback to be registered which is executed when a button press is needed? I think the second solution might be the neatest though I am open to suggestions.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+#: uses over MQTT -- and ``0x31`` are accepted onto the wire and silently
+#: dropped, so a wrong value here looks like a device that ignores the command.
+REALTIME_ROUTING = "21"
+
+#: ``a3[0]`` value emitted while a firmware update is in progress. It is absent
+#: from the published status enum, and a decoder treating the field as tri-state
+#: mis-renders every frame sent during an update.
+WORK_STATUS_UPDATING = 5
+
+BRIGHTNESS_VALUES = (1, 2, 3)
```
flip-dots, line 41: Does it not use 0 for off? That would allow for the [LightStatus](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/states.py#L71) enum to be used like it is for other devices.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+DISPLAY_TIMEOUT_VALUES = (60, 300, 1800)
```
flip-dots, line 42: Any reason the [DisplayTimeout](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/states.py#L106) enum cant be used?

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+        """
+        await self._send_command(
+            cmd=CMD_REALTIME,
+            parameters={
+                "a1": {"value": REALTIME_ROUTING},
+                "a2": {"type": 1, "value": (1).to_bytes(1)},
+            },
+        )
+
+    async def disable_realtime_telemetry(self) -> None:
```
flip-dots, line 193: What is the purpose of this?

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+            cmd=CMD_SUBSCRIBE,
+            parameters=SUBSCRIBE_PARAMETERS,
+        )
+        return KEEP_ALIVE_INTERVAL
+
+    #####################
+    # Realtime telemetry#
+    #####################
+
+    async def enable_realtime_telemetry(self) -> None:
```
flip-dots, line 178: Does the device care if it gets this more than once? If so I think it makes more sense to put this in the [_post_connect()](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/device.py#L238) for this model.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+
+    #: ``c490`` is a protobuf blob, walked into :attr:`summary` rather than the
+    #: flat TLV the other telemetry frames use.
+    _PROTOBUF_TELEMETRY_COMMANDS: tuple[str, ...] = ("c490",)
+
+    #: The c490 field decoding here is validated against this schema revision;
+    #: an older revision (the 2025 ``_0002``) or a newer one is warned about.
+    _VALIDATED_SUMMARY_SCHEMA: str = "charging_pps_series_c_0005"
+
+    async def _keep_alive(self) -> int | None:
```
flip-dots, line 154: The keep alive is only meant to be used for keeping the connection alive, not for polling telemetry. Polling for telemetry is meant to be done via [get_status_update](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/c1000.py#L463). Polling is currently done manually and at some point I do intend on making it so that polls are automatically done and telemetry is combined but that is something for a later PR.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+    async def enable_realtime_telemetry(self) -> None:
+        """Arm the realtime telemetry latch.
+
+        With this armed the device reports every change on its own, rather than
+        answering one reading per poll. The latch is persistent: it survives
+        disconnect, reconnect and session change, so it only needs arming once.
+        """
+        await self._send_command(
+            cmd=CMD_REALTIME,
+            parameters={
```
flip-dots, line 187: The parameters should be constants wherever possible.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+                "a1": {"value": REALTIME_ROUTING},
+                "a2": {"type": 1, "value": (0).to_bytes(1)},
+            },
+        )
+
+    ###################
+    # System settings #
+    ###################
+
+    async def turn_display_on(self) -> None:
```
flip-dots, line 220: Its a bit of a nit-pick but the rest of the [devices](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/c1000.py#L364) put commands (e.g AC on/off) after the properties (e.g battery percentage, time remaining).

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+        """Turn the display on."""
+        await self._send_command(cmd=CMD_SYSTEM, parameters=_parameters("a2", 1))
+
+    async def turn_display_off(self) -> None:
+        """Turn the display off."""
+        await self._send_command(cmd=CMD_SYSTEM, parameters=_parameters("a2", 0))
+
+    async def set_display_brightness(self, brightness: int) -> None:
```
flip-dots, line 228: The method name should be [set_display_mode](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/c1000.py#L412) and it should use the same interface.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+        """
+        if brightness not in BRIGHTNESS_VALUES:
+            raise ValueError(f"Brightness must be one of {BRIGHTNESS_VALUES}")
+
+        await self._send_command(
+            cmd=CMD_SYSTEM,
+            parameters=_parameters("a3", brightness),
+        )
+
+    async def set_display_timeout(self, seconds: int) -> None:
```
flip-dots, line 242: Needs to use the same interface as [set_display_timeout](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/c1000.py#L430)

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+        await self._send_command(
+            cmd=CMD_SYSTEM,
+            parameters=_parameters("a4", seconds, type_=2),
+        )
+
+    async def set_max_battery_percentage(self, percentage: int) -> None:
+        """Set the charge cap, above which the device stops charging.
+
+        The app only offers 80/85/90/95/100, but that is a UI convention -- the
+        firmware accepts any percentage, verified by setting 99 over BLE.
```
flip-dots, line 260: Technical details in method docstrings should be [notes](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-note) or warnings.

flip-dots: I think it should probably be a warning, someone might try and do something stupid otherwise (i.e set the charge cap to 0).

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+        """
+        if not 0 <= percentage <= MAX_PERCENTAGE:
+            raise ValueError(f"Percentage must be 0-{MAX_PERCENTAGE}")
+
+        await self._send_command(
+            cmd=CMD_SYSTEM,
+            parameters=_parameters("ab", percentage),
+        )
+
+    async def set_ac_output_timeout(self, seconds: int) -> None:
```
flip-dots, line 290: Needs to be [set_ac_timer](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/f2600.py#L394). Its important that parameters and commands all use the same names / interfaces where possible across devices, otherwise it leads to issues like [this](https://github.com/flip-dots/HaSolixBLE/issues/50).

As an aside at some point I might have to introduce some sort of check/rule for PRs that automatically checks that the interfaces are correct, since its easy for stuff to slip through the cracks.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+            disables the timer.
+        :raises ValueError: If the value is out of range or not a whole step.
+        """
+        _validate_timeout(seconds)
+        await self._send_command(
+            cmd=CMD_AC_OUTPUT,
+            parameters=_parameters("a3", seconds, type_=3),
+        )
+
+    async def set_dc_output_timeout(self, seconds: int) -> None:
```
flip-dots, line 305: Same as above issue with AC timer

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+        """Set the AC charging-power limit (W).
+
+        Read it back with :attr:`ac_input_limit`.
+
+        :param watts: AC charging power limit, 500-1800 W (the range the app
+            offers; sent as ``a4`` under the ``4101`` AC command).
+        :raises ValueError: If the value is out of range.
+        """
+        if not 500 <= watts <= 1800:
+            raise ValueError("AC charging power must be between 500 and 1800 W")
```
flip-dots, line 330: This is more a reminder to myself than anything, but with all these devices having different limits, I think it might be a good idea to have properties in the class for getting the upper and lower limits at some point.

I think thats best left for a later PR though which also sorts out the other devices which support this command.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+    @property
+    def ac_frequency(self) -> int:
+        """AC mains frequency (Hz).
+
+        :returns: Frequency in Hz, or default int value if there is no data.
+        """
+        return self._parse_int("a4", begin=7, end=8)
+
+    @property
+    def ac_input_limit(self) -> int:
```
flip-dots, line 445: Should be [ac_charging_power](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/f2600.py#L309).

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+
+        Distinct from :attr:`ac_input_limit`, which is the user-configured
+        limit; this is the fixed maximum the unit can draw.
+
+        :returns: Maximum input power in watts, or default int value if there
+            is no data.
+        """
+        return self._parse_int("a3", begin=5, end=7)
+
+    @property
```
flip-dots, line 464: Maybe just include this as a technical note in the battery_percentage property, I think its best to keep the interface simple where possible.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+        PortStatus.INPUT signifies the mains lead is present, NOT_CONNECTED that
+        it is absent. Presence only -- it does not imply current is flowing.
+
+        :returns: Status of the AC input.
+        """
+        return PortStatus.from_input_only(self._parse_int("a7", begin=4, end=5))
+
+    @property
+    def ac_output_timeout(self) -> int:
+        """AC output auto-off countdown (s), 0 when no timer is set.
```
flip-dots, line 490: Is this just a regular timer that you set or is this an idle timer? If its a regular timer it should be [ac_timer](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/f2600.py#L147) and there should be an [ac_timer_remaining](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/f2600.py#L139)

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+
+        :returns: Countdown in seconds, or default int value if there is no
+            data.
+        """
+        return self._parse_int("a4", begin=1, end=5)
+
+    @property
+    def dc_output_timeout(self) -> int:
```
flip-dots, line 498: Same comment as above

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+        """DC output auto-off countdown (s), 0 when no timer is set.
+
+        :returns: Countdown in seconds, or default int value if there is no
+            data.
+        """
+        return self._parse_int("a4", begin=9, end=13)
+
+    @property
+    def ac_output_mode(self) -> int:
```
flip-dots, line 507: Probably makes sense to re-use the [power_saving_mode_enabled](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/f2600.py#L329) property here.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+    @property
+    def dc_12v_output_mode(self) -> int:
+        """12 V DC output mode -- 0 normal, 1 smart (auto-off below 3 W).
+
+        :returns: Mode, or default int value if there is no data.
+        """
+        return self._parse_int("a4", begin=13, end=14)
+
+    @property
+    def device_timeout_minutes(self) -> int:
```
flip-dots, line 523: Should be [device_timeout](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/c300dc.py#L358)

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+        :returns: True if enabled, else False, or default bool value if there
+            is no data.
+        """
+        if self._data is None:
+            return DEFAULT_METADATA_BOOL
+
+        return bool(self._parse_int("a4", begin=23, end=24))
+
+    @property
+    def display_on(self) -> bool:
```
flip-dots, line 556: The naming scheme is a tad inconsistent here but the existing convention for this particular thing is [is_display_on](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/f2600.py#L341), though display_on kind of makes more sense, I might end up changing the existing ones in a later PR but for now this needs to use the existing scheme.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+        :returns: True if lit, else False, or default bool value if there is no
+            data.
+        """
+        if self._data is None:
+            return DEFAULT_METADATA_BOOL
+
+        return bool(self._parse_int("a4", begin=22, end=23))
+
+    @property
+    def display_brightness(self) -> int:
```
flip-dots, line 568: Should be [display_mode](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/f2600.py#L353)

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+
+        :param index: Slot index, 0-6.
+        :returns: Dotted version string, or default str value if there is no
+            data or the block is short.
+        """
+        if self._data is None or "f9" not in self._data:
+            return DEFAULT_METADATA_STRING
+
+        # value_legacy retains the leading type byte.
+        block = self._data["f9"].value_legacy[1:]
```
flip-dots, line 598: You can use .value instead of .value_legacy here.

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+    @property
+    def software_version(self) -> str:
+        """Main software version.
+
+        :returns: Firmware version or default str value.
+        """
+        return self._version_slot(0)
+
+    @property
+    def software_version_sub_mcu(self) -> str:
```
flip-dots, line 614: Does it make more sense to use [software_version_controller](https://github.com/flip-dots/SolixBLE/blob/e91c2527cecf0d8635b72dc16a79d65d750e48f2/SolixBLE/devices/c1000.py#L279) here?

SolixBLE/devices/c2000g2.py
```diff
@@ -0,0 +1,761 @@
...
+        # value_legacy retains the leading type byte (0x04, a `bin` field).
+        block = self._data["c0"].value_legacy[1:]
+        if not block:
+            return None
+
+        fields = block[1 + block[0] :]
+        return fields if len(fields) >= SUB_PACKAGE_TAIL_LENGTH else None
+
+    @property
+    def expansion_present(self) -> bool:
```
flip-dots, line 676: I think it makes more sense to move this into num_expansion, since num_expansion is used by other devices.

flip-dots (commented):
So I have taken a look at the docs changes and the C2000G2 changes and I don’t see any major issues with them but I am out of time to review this for this weekend so have not taken a look at the rest, though I plan on doing so next weekend.

My overall thoughts so far:
- Multiple PRs is better but still kind of difficult to review since comments and their fixes are spread across multiple PRs, I recently started using [GitHubs stacked PRs](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests) for this kind of thing and that helps, you might want to consider using them or else keep PRs draft until all of its dependencies are merged.
-  This needs to be split up more
   - Support for the C2000G2 and devices which need button presses for pairing need to be a separate PR
   - Support for dynamically determining the negotiation path needs to be its own PR
   - Schedule support for the 250w charger needs to be its own PR
   - Support for the 240w charger needs to be its own PR
   - Support for using push and poll telemetry needs to be its own PR
- Be careful with method names/interfaces, they need to conform to the existing patterns.

kb1ibt:
>   I recently started using [GitHubs stacked PRs](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests) for this kind of thing and that helps, you might want to consider using them or else keep PRs draft until all of its dependencies are merged.


The problem with stacks is that it also needs push permission, so when I tried, I got:

```
Checking stack state...
Pushing to https://github.com/flip-dots/SolixBLE.git...
✗ failed to push pr/negotiation-path: failed to run git: remote: Permission to flip-dots/SolixBLE.git denied to kb1ibt.
fatal: unable to access 'https://github.com/flip-dots/SolixBLE.git/': The requested URL returned error: 403
```

flip-dots:
> > I recently started using [GitHubs stacked PRs](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests) for this kind of thing and that helps, you might want to consider using them or else keep PRs draft until all of its dependencies are merged.
> 
> The problem with stacks is that it also needs push permission, so when I tried, I got:
> 
> ```
> Checking stack state...
> Pushing to https://github.com/flip-dots/SolixBLE.git...
> ✗ failed to push pr/negotiation-path: failed to run git: remote: Permission to flip-dots/SolixBLE.git denied to kb1ibt.
> fatal: unable to access 'https://github.com/flip-dots/SolixBLE.git/': The requested URL returned error: 403
> ```

That sucks, it looks like its not supported for merging across forks, I hope they fix that because stacks are nice af. For now if you could keep stuff as draft when its not ready / dependencies are not merged that would make things easier for me, I am but one man, and unfortunately not a particularly well organised one :)

kb1ibt:
Rebuilt as seven PRs from `main`, split the way you listed: #70 negotiation path, #71 token + button pairing, #72 C2000 G2 (with the c490 decode and the subscribe fix that overlaps #67), #73 250W snapshot + schedule/timer, #74 250W push-and-poll, #75 A91B2, and #76 moving the shared Gen 2 surface down to the C1000 G2 for the #22 owners. Names and interfaces follow the existing devices; each inline thread here has a reply saying where its fix landed.

On the fork/stacking point from above: agreed, so I've kept it to plain dependent PRs and done exactly what you asked — everything stays a draft until its dependencies land. Only #70 is ready now; each dependent PR is a draft that names the PR it depends on, and I flip it to ready (rebased onto `main`) once that parent merges. The tree is #70 → #71 → {#72, #73}, #72 → #76, and #73 → {#74, #75}, so it's one layer at a time: #70 is the only thing that needs a look right now, and nothing downstream is ready for you until it lands.

And genuinely, no worries on the organisation front — I've got ADHD, so I need all of the helps I can get.

Closing this one, #49 and #50 in favour of those.

PR #52, opened 2026-07-21
docs: C2000 G2 + A91B2 station pages, owner_user_id, refreshed support tables

kb1ibt:
Split out of #45 per review — the docs.

Adds device pages for the **C2000 G2** and the **A91B2 charging station**, an `owner_user_id` guide, and refreshes the support tables/index for the new devices.

**Stacked on #51** (Prime devices) → #50 (c490) → #49 (negotiation) → #48 (reassembly). This is the final piece; with all five merged, the tree equals the original #45. Docs-only.

PR #54, opened 2026-07-23
feat(prime): re-arm realtime telemetry with a keepalive loop

kb1ibt:
## What

Prime chargers stream telemetry ~1/s after a `420b` realtime trigger, but the realtime window lapses after ~10 s — so the stream stalls unless the client keeps re-arming it. This adds a generic keepalive task to `SolixBLEDevice`:

- When a subclass sets `_KEEPALIVE_CMD`, `connect()` starts a background loop that re-sends it every `_KEEPALIVE_INTERVAL` (default 8 s, ~1 s under the device window), and `disconnect()` cancels it.
- `PrimeDevice` sets `_KEEPALIVE_CMD = "420b"` (bare `a10121` payload via `_send_command`, so it carries the live session timestamp).
- `PrimeChargingStation240w` (A91B2, CBC lineage) overrides `_keepalive_loop` to send `420b` as `a10121fe0503<ts>` through `_send_packet` (`response=True`), matching its own `_post_connect` framing.

Without this, a client that periodically pokes the device for a refresh ends up re-running `_post_connect` (re-sending one-time confer commands) and adding a second writer to the command characteristic — the device answers, then drops the link. The keepalive loop makes the device stream on a single steady writer and keeps telemetry flowing indefinitely.

## Stacked on #48–#52

This branch is stacked on the still-open Prime / telemetry PRs (**#48–#52**). Until those land on `main`, the diff here shows their commits too — the **only new commit in this PR is the top one**, `feat(prime): re-arm realtime telemetry with a keepalive loop` (`5e31f9d`). Please review/merge the stack first; this is marked draft until then.

## Hardware validation

- **A2345** (Prime 250 W charger, GCM) — base `PrimeDevice` `420b` keepalive loop; stream sustained indefinitely.
- **A91B2** (Prime 240 W charging station, CBC) — the `_keepalive_loop` override; validated live across multiple sessions (continuous `4303` stream + `4a0b` acks at ~8 s, no drops, no confer re-sends).

⚠️ The **160 W charger** and **power bank** inherit `PrimeDevice._KEEPALIVE_CMD` but were **not** hardware-tested — they will now run the base `420b` keepalive loop untested. Flagging for maintainer awareness.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kb1ibt:
Folding this into #51 instead — the keepalive is what makes that PR's live telemetry actually keep streaming, so it belongs with the device classes rather than as a separate PR on top. Both commits (the 420b keepalive machinery and a follow-up fixing the station's hardcoded confer timezone) are now in `pr/prime`, and #52 is rebased on top. Closing; no work is lost.

Issue #55, opened 2026-07-27
Title: 767 PowerHouse (1st gen, E8:EE:CC:3F:A6:1F) uses different GATT profile than expected — BleakCharacteristicNotFoundError on connect


chriba2567:
**Environment**
- Device: Anker PowerHouse 767, 1st generation (Bluetooth-only, no WiFi)
- Broadcast name: `767_PowerHouse`
- MAC: `E8:EE:CC:3F:A6:1F`
- HA 2026.7.2 / HaSolixBLE 3.5.0 / SolixBLE 3.8.0
- Tested via both an ESPHome active Bluetooth proxy (ESP32, -57 dBm) and
  a local BlueZ adapter (-73 dBm) — identical result on both, so this is
  not a range/proxy issue.

**Problem**

The integration correctly discovers the device (after patching the
`manifest.json` bluetooth matcher — see below) and lets me pick model
"F2000 (767)". Confirming the setup fails with the UI showing
"Failed to negotiate encryption", which turned out to be misleading —
the actual error is a `BleakCharacteristicNotFoundError`:

Logger: SolixBLE.device
Error subscribing/negotiating with '767_PowerHouse'!

Traceback (most recent call last):
File "SolixBLE/device.py", line 156, in connect
await self._client.start_notify(
UUID_TELEMETRY, partial(self._process_notification, self._client)
)
File "bleak/init.py", line 892, in start_notify
characteristic = _resolve_characteristic(char_specifier, self.services)
File "bleak/init.py", line 432, in _resolve_characteristic
raise BleakCharacteristicNotFoundError(char_specifier)
bleak.exc.BleakCharacteristicNotFoundError:
Characteristic 8c850003-0302-41c5-b46e-cf057c562025 was not found!

ESPHome BLE client logs confirm the connection, service discovery, and
MTU negotiation (247) all complete cleanly with no security/pairing
events at all — the disconnect is client-initiated, ~50-110ms after
service discovery completes, consistent with the exception above (the
library gives up right after failing to find the expected
characteristic, not because of an actual encryption/pairing failure).

**Actual GATT structure (verified via nRF Connect on iOS)**

| | Expected by SolixBLE (F2000 profile) | Actual on this device |
|---|---|---|
| Advertised / primary service | `0000ff09-0000-1000-8000-00805f9b34fb` (initial matcher) | `014bf5da-0000-1000-8000-00805f9b34fb` |
[...]

PR #64, opened 2026-08-21
Add support for older F2000/767 firmware version

impala454:
This PR adds in support for the older model F2000/767 firmware.  This firmware is not upgradable to the latest and has a completely different scheme with no handshakes or negotiation.  This PR adds almost all of the functionality the Anker app has to offer to this library.  The only minor edit to the base class was just to allow for the different UUIDs needed and still maintain inheritance.


flip-dots:
The new constructs in #61 should be able to make the commands and bit mangling a lot cleaner, with any luck that will be merged soon and it will be possible to take advantage of it.

impala454:
I took a look at #61 and seems like a good change but this older firmware is so completely different I don't think it would  change this PR.  There is no negotation, no multi part packets, and the parsing is fixed fields, so I would probably still be overriding ` _process_notification` anyways.  The only thing I needed to touch in the base was just to allow the different UUID to be recognized.  I don't think that would change with #61 merged either.  Let me know what you think.

impala454:
btw this closes #55

flip-dots:
Wow..., I only skimmed this before but this is very different, its practically an entirely different protocol (I think we are now on the 4th or 5th protocol at this point). I would like to see some tests for this before I would be happy to merge it for things like sending commands and parsing telemetry. Without any tests or test data I cant properly review this or be sure that it wont be broken by future updates to this library, its especially important given that I don't have that model to physically test with.

impala454:
Totally agreed, I'll get a good suite of tests going and let you know when to look again.

impala454:
@flip-dots should be good to go now- once this is in I'll test the HA side repo for a bit w/this then get that one merged also.

flip-dots:
I have re-written it to better fit in with the existing code structure and fixed some other issues (functions/properties with inconsistent names), could you confirm that the new code works, also any suggestions are welcome. I ended up removing the parsing of acknowledgements since they are a bit redundant given that a full status update is requested after a command is sent, feel free to have a go at re-adding them if you want though.

impala454:
Took a short look and can test later.  Totally get the "legacy" rename that makes more sense.  The acks are important because they only occur when the user presses the physical butttons (I thought I had explained that in the comments).  Without them in there the device can get out of state when physical buttons are pressed.  This is also one of the reasons I chose to implement it in a "parse on receipt" instead of "parse on demand" way.  I think your parsers are fine but they're omitting a lot of the work that was done to actually figure out what the packet structure was (various parts of the header meaning different types).  Instead you've turned it into matching hard coded hex values as strings which can be very brittle.  Also keeping the telemetry class and parsing of the packet independent of the HA getter functions meant it was a lot more readable and editable and keeps the "business logic" independent of the external API.  No worries if we don't see eye to eye just those are my initial thoughts on the changes.

